PR #12, opened 2019-12-03
Remove public from interface property

davidhenley:


JasonGT:
Awesome, thanks @davidhenley. 👍

PR #19, opened 2019-12-08
fix path \Src->\src

mfehlhaber:
Had to fix capitalization of src directory in order to be able to build on Linux.

```
grep -rl --include *.sln --include *.csproj "Src" . |xargs sed -i 's/Src/src/g'
```


JasonGT:
Thanks for sorting that out @mfehlhaber. 👍

PR #20, opened 2019-12-08
Small improvement to ValidationException ctor.

parc1vaL:
This reduces the iterations in the constructor of ValidationException to the necessary minimum.


JasonGT:
I love it! So much cleaner. Thanks! 👍

PR #28, opened 2019-12-12
Make AuditableEntity abstract

davidhenley:


jasontaylordev:
Thanks again @davidhenley! 👍

PR #36, opened 2019-12-16
Using more general type (IEnumerable) as input parameter

MoienTajik:
I've changed List to IEnumerable according to ["**Parameter Design**"](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/parameter-design) section of "**Microsoft Design Guidelines**":

> ✔ **DO** use the least derived parameter type that provides the functionality required by the member.
> 
> For example, suppose you want to design a method that enumerates a collection and prints each item to the console. Such a method should take IEnumerable as the parameter, not ArrayList or IList, for example.


jasontaylordev:
Thank you 👍

PR #54, opened 2020-01-17
Fixed issue where if an Controller or Controller method was set to an…

taylorc:
If a controller is set to be anonymous or has the authorise attribute removed the call of that API method would result in an exception


jasontaylordev:
Thanks, @taylorc! 👍

PR #73, opened 2020-02-21
Added password config to remove digit requirement. Fixes #68.

### LINKED ISSUE #68: [Bug] Seed data for adding default user is not updating database
**Describe the bug**
Title says it all. `ApplicationDbContextSeed` is running but the default user is not getting added to the Db. I can register new users fine; it's just the seed data that's the issue.

**To Reproduce**
1. Check out a fresh copy of `master`.
2. Delete existing `CleanArchitectureDb` if present.
3. Run application from VS.
4. Verify that `dbo.AspNetUsers` is empty.

**Expected behavior**
`dbo.AspNetUsers` should contain "jason@clean-architecture" user.

**Additional context**
The seed logic is the same as eShopOnWeb. I've tested that solution too and it seeds the users fine. My guess is some config has been missed in start-up.

GFoley83:
Fixes a bug in `ApplicationDbContextSeed` where adding a user with a password that violates the password requirements silently fails and does not add the user to the Db.

### DIFF
--- a/src/Infrastructure/DependencyInjection.cs
+++ b/src/Infrastructure/DependencyInjection.cs
@@ -29,8 +29,16 @@ namespace CleanArchitecture.Infrastructure
 
             services.AddScoped<IApplicationDbContext>(provider => provider.GetService<ApplicationDbContext>());
 
-            services.AddDefaultIdentity<ApplicationUser>()
-                .AddEntityFrameworkStores<ApplicationDbContext>();
+            services.AddDefaultIdentity<ApplicationUser>(options =>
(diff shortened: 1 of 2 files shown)


jasontaylordev:
Hi @GFoley83 - thanks for the contribution. Rather than change the defaults I opted to just increase the complexity of the default user password - see https://github.com/jasontaylordev/CleanArchitecture/commit/fe4caf6286cc4f5a6c71621733c01315a6aebc6f.

Regardless, thank you very much for highlighting this issue. 👍

PR #75, opened 2020-02-23
Add some steps to the Getting Started Section

MarcosMeli:
Node Js link, explain that the template will use current directory as a project name and namespace


jasontaylordev:
Thank you 👍

PR #97, opened 2020-03-29
Fixed example Migration line in 'README.md'

engineers-tools:
The --output-dir option was missing on the example line provided under the Database Migrations paragraph. If a user copies and pastes this directly, it will save the new migration in a different directory causing conflicts with the migrations already in the template


jasontaylordev:
Thanks 😀

engineers-tools:
No worries

Issue #106, opened 2020-04-21
[Feature request] Angular interceptor


giovanellij:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. 

I'm always frustrated when I have to handle each error in the angular application.

**Describe the solution you'd like**
I would like to have an interceptor to handle all the errors and only worry about the ok response.

**Additional context**
I have tried to make this work but the catchError does not catch the response body.

jasontaylordev:
This is a good idea. Take a look at this resource:

- [Expecting the Unexpected — Best practices for Error handling in Angular](https://medium.com/angular-in-depth/expecting-the-unexpected-best-practices-for-error-handling-in-angular-21c3662ef9e4)

It outlines a comprehensive approach. What do you think?

jasontaylordev:
I think there's already a lot of good guidance on this topic, such as the article above. 

Closing.

PR #115, opened 2020-04-26
Add frontend interceptors

Greenmachine52:
#106  adds functial mentioned in issue.

Interceptors are added for both frontend code in general and httpresponses.
Two handlers in ClientApp\src\error-handling and registered in main module.
They should be self explanitory.

Cheers!

### DIFF
--- a/src/WebUI/ClientApp/src/app/app.module.ts
+++ b/src/WebUI/ClientApp/src/app/app.module.ts
@@ -14,6 +14,8 @@ import { TodoComponent } from './todo/todo.component';
 import { ApiAuthorizationModule } from 'src/api-authorization/api-authorization.module';
 import { AuthorizeGuard } from 'src/api-authorization/authorize.guard';
 import { AuthorizeInterceptor } from 'src/api-authorization/authorize.interceptor';
+import { GlobalErrorHandler } from 'src/error-handling/global-error.handler';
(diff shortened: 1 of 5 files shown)


MCCshreyas:
Hi, @Greenmachine52 Seems like you just added boilerplate for interceptors. Why don't you add some values to `ServerErrorInterceptor`  like displaying bootstrap popup modal showing error information that will help end-user to fix the corresponding issue !?

Greenmachine52:
@MCCshreyas  Hi there! I appreciate the feedback but fail to agree with you at this point.

What I added is a working template for catching errors and a architectural layer. So a comment or just logging the error into console should be enough.

Adding a popup from bootstrap or any other styling framework isn't something I'd like to see since anyone actually using this will probobly have significant changes to the UI on his or her own.

Integrating any logging or notification mechanism with this boilerplate should be a matter or minutes.

Howeever, I am open to discuss further on or implement after a second opinion.

MCCshreyas:
It's just a template so the user can modify the behaviour as per their comfort for sure. 
What I noticed is that, the interceptors right now just empty, and they are nothing doing but just residing in the request, response pipeline. So I think it's better to have something helpful there rather than being empty. 
After all, they are Ok.

jasontaylordev:
Apologies for taking so long to respond. This is a good approach but I'd agree with @MCCshreyas, better to have a sample rather than empty implementations. Closing, but thanks anyway.

PR #121, opened 2020-04-28
Added pagination support for query responses

MCCshreyas:
This PR will add support for pagination for query responses. As pagination has special importance in building enterprise applications and small applications, as no one wants to get their client flooded with 1000's of database records in one go.

Implementing pagination is so simple with the following approach. Please see FAQ's section for more clarification.

For demonstration purposes, I have created a new action and query in the application as I don't want to break existing action to break because of pagination as it requires changes in frontend as well.

Note - Please see PR file diff for changes.

## FAQ

1.  **How can I add pagination support for my queries**

    The following are a couple of steps that you should follow.
     
    1.  Inherit your query from `PaginationQuery` so that you can set the `PageNumber` and `PageSize` parameter which are important in pagination and can be used with LINQ to get the corresponding result.

    _Query_

    ```csharp
    public class GetTodosByListIdQuery : PaginationQuery, IRequest<List<TodoItemResultRecord>>
    {
        public int ListId { get; set; }
    }
    ```

    _Query handler_

    ```csharp
    public class GetTodosByListIdQueryHandler : IRequestHandler<GetTodosByListIdQuery, List<TodoItemResultRecord>>
      {
          private readonly IApplicationDbContext _context;
          private readonly IMapper _mapper;

          public GetTodosByListIdQueryHandler(IApplicationDbContext context, IMapper mapper)
          {
              _context = context;
              _mapper = mapper;
          }

          async Task<List<TodoItemResultRecord>> IRequestHandler<GetTodosByListIdQuery, List<TodoItemResultRecord>>.Handle(GetTodosByListIdQuery request, CancellationToken cancellationToken)
          {
              if(request.SkipPageCount >= 0)
              {
                  return await _context.TodoItems.Skip(request.SkipPageCount).Take(request.PageSize.GetValueOrDefault())
                      .Where(t => t.ListId == request.ListId)
                      .ProjectTo<TodoItemResultRecord>(_mapper.ConfigurationProvider)
                      .ToListAsync(cancellationToken);
              }

              var records = await _context.TodoItems
                      .Where(t => t.ListId == request.ListId)
                      .ProjectTo<TodoItemResultRecord>(_mapper.ConfigurationProvider)
                      .ToListAsync(cancellationToken);

              return records;
          }
    ```

    As you can see above we have already calculated the `SkipPageCount` value for you based upon the `PageNumber` and `PageSize` parameter, so you have to just use that in your Ef core LINQ query to get the corresponding result.

    3. And finally, you have to tweak your corresponding controller action definition so that it can take those parameters from users like below. And also decorate the controller with `PagedResponse` attribute. 

    ```csharp
      [PagedResponse]
      [HttpGet("result/{id}")]
      public async Task<List<TodoItemResultRecord>> GetTodos([FromQuery] PaginationQuery paginationQuery ,int id)
      {
          var result = await Mediator.Send(new GetTodosByListIdQuery
          {
              ListId = id,
              PageNumber = paginationQuery.PageNumber,
              PageSize = paginationQuery.PageSize
          });
          return result;
      }
    ```

    By doing all of the above steps the pagination will start working for queries. You will get an API
    response something like below.
image

    As you can see above, the response also contains links to the **Next and Previous page**, so that any front end application can take advantage of it for navigation without doing any kind of complex calculation at the front end for evaluating links.

So in this way, one can implement pagination for their queries. As always I am listening to community members for feedback and open for changes for the betterment of implementation.

Thanks,

### DIFF
--- a/src/Application/Common/Models/PaginatedResponse.cs
+++ b/src/Application/Common/Models/PaginatedResponse.cs
@@ -1,0 +1,12 @@
+﻿namespace CleanArchitecture.Application.Common.Models
+{
+    public class PaginationResponse
+    {
+        public object Data { get; set; }
+        public int? PageNumber { get; set; }
+        public int? PageSize { get; set; }
+
+        public string NextPage { get; set; }
+        public string PreviousPage { get; set; }
+    }
+}
--- a/src/Application/Common/Models/PaginationQuery.cs
+++ b/src/Application/Common/Models/PaginationQuery.cs
@@ -1,0 +1,10 @@
+﻿namespace CleanArchitecture.Application.Common.Models
+{
+    public class PaginationQuery
+    {
+        public int PageNumber { get; set; } = 0;
+        public int? PageSize { get; set; } = 10;
+
+        public int SkipPageCount => (PageNumber - 1) * PageSize.GetValueOrDefault();
+    }
+}
--- a/src/Application/TodoLists/Queries/GetTodosByListId/GetTodosByListIdQuery.cs
+++ b/src/Application/TodoLists/Queries/GetTodosByListId/GetTodosByListIdQuery.cs
@@ -1,0 +1,48 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using AutoMapper;
+using AutoMapper.QueryableExtensions;
+using CleanArchitecture.Application.Common.Interfaces;
+using CleanArchitecture.Application.Common.Models;
+using MediatR;
+using Microsoft.EntityFrameworkCore;
+
+namespace CleanArchitecture.Application.TodoLists.Queries.GetTodosByListId
+{
+    public class GetTodosByListIdQuery : PaginationQuery, IRequest<List<TodoItemResultRecord>>
+    {
+        public int ListId { get; set; }
+    }
+
+    public class GetTodosByListIdQueryHandler : IRequestHandler<GetTodosByListIdQuery, List<TodoItemResultRecord>>
+    {
+        private readonly IApplicationDbContext _context;
+        private readonly IMapper _mapper;
+
+        public GetTodosByListIdQueryHandler(IApplicationDbContext context, IMapper mapper)
+        {
+            _context = context;
+            _mapper = mapper;
+        }
+
+        async Task<List<TodoItemResultRecord>> IRequestHandler<GetTodosByListIdQuery, List<TodoItemResultRecord>>.Handle(GetTodosByListIdQuery request, CancellationToken cancellationToken)
+        {
+            if(request.SkipPageCount >= 0)
+            {
+                return await _context.TodoItems.Skip(request.SkipPageCount).Take(request.PageSize.GetValueOrDefault())
+                    .Where(t => t.ListId == request.ListId)
+                    .ProjectTo<TodoItemResultRecord>(_mapper.ConfigurationProvider)
+                    .ToListAsync(cancellationToken);
+            }
+
+            var records = await _context.TodoItems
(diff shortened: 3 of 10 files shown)


ShamalFdo129:
@MCCshreyas 
I was going through your solution, this is a nice solution for server side pagination. do we want to send total number of records for rendering UI components in Angular?

MCCshreyas:
@ShamalFdo129 Hi, Thanks for your response. 
>do we want to send the total number of records for rendering UI components in Angular?

To answer your question, I think it really depends on what we are showing to the user. If your use case is that you show the total number of records on UI along with the page control then you can add a new property to the pagination model and can hold the total number of records in it and then you can use that property in your front end to show the total number of records to the user.

ShamalFdo129:
yeap it depends on the requirement. I have to show total number of records and render pagination UI controls based on total number of record. 
Thank you for your response.

jasontaylordev:
Hi @MCCshreyas - thanks again for your patience. I appreciate that you've put a lot of effort into this pull request, I love the documentation - we should start building out the wiki. I've reviewed your approach to pagination, and while it is very good, it's not the approach I'd like to take in this template. Some feedback:

- As mentioned above it's missing a total number of records. It's true that depending on the requirements you may not need this, but I think the general use case is that it would be required, in order to set correct states for navigation, e.g. https://getbootstrap.com/docs/4.0/components/pagination/.
- While `PaginationQuery` does cut down on the code, I would rather just list the properties in the query itself. I think this will be simpler to manage in the long run. It's also a little more explicit about the required inputs.
- The PagedResponse attribute is in WebUI and dependent on ASP.NET Core. Prefer to keep all of the pagination logic inside of Core.

I've reviewed some other approaches to pagination. The best one I can find is this one:

```csharp
   model.Results = await students
      .ProjectTo<Model>(_configuration)
      .PaginatedListAsync(pageNumber, pageSize);
```

This is from @jbogard's Contoso University sample. Recommend cloning and reviewing the implementation.

Hope that this is helpful feedback, feel free to disagree. Let me know what you think.

iayti:
> Hi @MCCshreyas - thanks again for your patience. I appreciate that you've put a lot of effort into this pull request, I love the documentation - we should start building out the wiki. I've reviewed your approach to pagination, and while it is very good, it's not the approach I'd like to take in this template. Some feedback:
> 
> * As mentioned above it's missing a total number of records. It's true that depending on the requirements you may not need this, but I think the general use case is that it would be required, in order to set correct states for navigation, e.g. https://getbootstrap.com/docs/4.0/components/pagination/.
> * While `PaginationQuery` does cut down on the code, I would rather just list the properties in the query itself. I think this will be simpler to manage in the long run. It's also a little more explicit about the required inputs.
> * The PagedResponse attribute is in WebUI and dependent on ASP.NET Core. Prefer to keep all of the pagination logic inside of Core.
> 
> I've reviewed some other approaches to pagination. The best one I can find is this one:
> 
> ```cs
>    model.Results = await students
>       .ProjectTo<Model>(_configuration)
>       .PaginatedListAsync(pageNumber, pageSize);
> ```
> 
> This is from @jbogard's Contoso University sample. Recommend cloning and reviewing the implementation.
> 
> Hope that this is helpful feedback, feel free to disagree. Let me know what you think.

I implement your advice. Pull Request #227. Can you check my solution @MCCshreyas. I think we must also add generic ServiceResponse for the project.

Hope that this is helpful. Have a nice coding.

MCCshreyas:
Thanks @jasontaylordev. Thanks for review comments. Am quite busy in this week, so not sure when I will do those changes to my PR. 🤕🤕

You can continue with @iayti PR implementation.

jasontaylordev:
No worries @MCCshreyas - thanks again for your contribution.

PR #129, opened 2020-05-07
Updated validation pipeline to execute async validations asynchronously

### LINKED ISSUE #128: Asynchronous validations run synchronously when invoked with Validate() instead of ValidateAsync()
Asynchronous validations must be invoked with **ValidateAsync()** [ValidationBehavior.cs](https://github.com/jasontaylordev/CleanArchitecture/blob/master/src/Application/Common/Behaviours/ValidationBehavior.cs) to be run asynchronously according to the [FluentValidation documentation](https://docs.fluentvalidation.net/en/latest/async.html)

> Warning
> If your validator contains asynchronous validators or asynchronous conditions, it’s important that you always call ValidateAsync on your validator and never Validate. If you call Validate, then your asynchronous rules will be run synchronously, which is not desirable.
> You should not use asynchronous rules when using automatic validation with ASP.NET as ASP.NET’s validation pipeline is not asynchronous. If you use asynchronous rules with ASP.NET’s automatic validation, they will always be run synchronously.

A fix requires to a change from calling Validate() to ValidateAsync() in the ValidationBehavior.cs

michalkrzych:
#128

### DIFF
--- a/src/Application/Common/Behaviours/ValidationBehavior.cs
+++ b/src/Application/Common/Behaviours/ValidationBehavior.cs
@@ -18,25 +18,19 @@ namespace CleanArchitecture.Application.Common.Behaviours
             _validators = validators;
         }
 
-        public Task<TResponse> Handle(TRequest request, CancellationToken cancellationToken, RequestHandlerDelegate<TResponse> next)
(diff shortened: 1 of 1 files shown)


jasontaylordev:
So much better! Thanks, @michalkrzych. 😊

PR #148, opened 2020-05-25
Throw on user creation failure and make find user more generic

### LINKED ISSUE #147: User creation in Testing could fail
The current code isn't handling failure scenarios:

```csharp
 public static async Task<string> RunAsUserAsync(string userName, string password)
    {
        using var scope = _scopeFactory.CreateScope();

        var userManager = scope.ServiceProvider.GetService<UserManager<ApplicationUser>>();

        var user = new ApplicationUser { UserName = userName, Email = userName };

        var result = await userManager.CreateAsync(user, password);

        _currentUserId = user.Id;

        return _currentUserId;
    }
```

I'll push a PR for the fix now.

AlQuraian:
Fixes #147 

Thanks a lot @jasontaylordev for the recent improvements to testing.
I think it's much better now!

### DIFF
--- a/tests/Applicaton.IntegrationTests/Testing.cs
+++ b/tests/Applicaton.IntegrationTests/Testing.cs
@@ -11,13 +11,14 @@ using Microsoft.Extensions.DependencyInjection;
 using Moq;
 using NUnit.Framework;
 using Respawn;
+using System;
 using System.IO;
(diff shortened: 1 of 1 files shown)


jasontaylordev:
Awesome, thank you 👍

PR #159, opened 2020-06-04
Remove constructor and use collection initializer

hsalameh:
It's better to initialize the collection in the declaration so that if more constructors were added, there would be no need to initialize the collection in every constructor.


jasontaylordev:
I agree - thanks. 👍

Issue #125, opened 2020-05-02
IMapTo?


sweeperq:
What do you think about adding an `IMapTo` interface that is basically the inverse of `IMapFrom`?

I implemented it in one of my projects and found it particularly useful when mapping simple Create/Update commands to Domain entities. Instead of manually mapping properties, or using `IMapFrom` and implementing a custom `Mapping` function on the command, it was nice to rely on the default `IMapTo` interface implementation for simple Command-to-Entity mappings.

I was starting the project from scratch, so when I implemented the feature, I renamed the `IMapFrom` `Mapping` function to `MapFrom`, and created a new `IMapTo` interface with a default `MapTo` function. I copied what you had in the MappingProfile and updated it to work for both interfaces.

Inject `IMapper` into your request handlers and you are all set to AutoMap the Commands to Entities.

jasontaylordev:
It sounds like a good approach, although I'd consider using an [explicit interface implementation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation). This would mean you would implement `IMapFrom.Mapping` and `IMapTo.Mapping` within the DTO. For your system, do you typically map both ways? If so you could just use one interface and perhaps utilise [reverse mapping](https://docs.automapper.org/en/stable/Reverse-Mapping-and-Unflattening.html) for simple scenarios.

I tend not to map from DTO to an entity as you can see in the template. I like to be in full control of the assignments that are being applied to the underlying domain.  With the manual approach, if new fields are added it's up to the developer to decide how and when they can be changed.

@brendan-ssw - we talk about this a bit - anything to add?

Thanks for raising the issue and please share your thoughts. 😃

Marusyk:
It is very interesting. I need `IMapTo` too
@jasontaylordev @sweeperq are there any code samples?

jasontaylordev:
Hi @Marusyk - no examples sorry. But if you review the logic for IMapFrom and MappingProfile (both in Application) I am sure you can figure it out. Also, you're in luck, someone recently contributed a pull request on this topic - https://github.com/jasontaylordev/CleanArchitecture/pull/167.

In relation to ignoring properties, check out the AutoMapper docs; https://docs.automapper.org/en/stable/Attribute-mapping.html?highlight=Ignore#ignoring-members.

Hope this helps.

jeevasusej:
Thank you for discussing about my pull request. Do you have any idea to merge my pull request?

PR #167, opened 2020-06-18
Automapping - IMapTo - IHaveCustomMappings

jeevasusej:
Automapper configuration for IMapTo and CustomMappings.


jasontaylordev:
Hi @jeevasusej - thanks for the contribution - apologies for taking so long to respond. As per https://github.com/jasontaylordev/CleanArchitecture/issues/125#issuecomment-625785416 I'd prefer not to have reverse mappings at this stage. I like to be explicit when updating entities.

PR #168, opened 2020-06-20
Changed HttpContextAccessor to retrieve context when getter is called

xontab:
Since `HttpContext` is using `AsyncLocal` internal while the `CurrentUserService` is using DI (scope), there might be cases where there is a synchronisation issue that causes the wrong or `null` UserId to be returned.  Is it safe to have a Singleton service which retrieves the current context from the accessor as suggested by David Fowler - https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AspNetCoreGuidance.md#do-not-store-ihttpcontextaccessorhttpcontext-in-a-field


jasontaylordev:
Awesome, thank you 👍

Issue #176, opened 2020-06-27
Mixed casing for the generated NSwag client filepath, breaks on Linux machines


MJM1989:
**Describe the bug**
The filepath specified as output in the nswag.json specification uses PascalCase while the import statements in controllers such as the TodoComponent and the FetchDataComponent uses lowercase. On a Linux machine, with a case-sensitive filesystem, this breaks. 

**To Reproduce**
1. Go to a Linux machine ;)
2. Create a directory with PascalCase (or camelCase), e.g. 'MyTestProject'
2. Initialize the template using 'dotnet new ca-sln'
3. Startup the project
4. In the browser, the following error message is displayed: "Cannot GET /"
5. In the console we see the following error appear: "ERROR in src/app/fetch-data/fetch-data.component.ts:2:56 - error TS2307: Cannot find module '../cleanmytestproject-api'."

**Expected behavior**
Initialiting a new solution from the template, should create a lowercase output file (e.g. 'cleanmytestproject-api') for the nswag.json, in conformance with the internal code style of the Angular project.

MJM1989:
Since my first reply is a bug report :S let me just start a comment with saying:
Thank you, and thanks to all involved, for creating such a nice template, full of many great ideas. 

I'm really looking forward to looking into this more carefully and hope I can help out by quickly making a PR for this issue ;)

jasontaylordev:
Thanks again for your PR. 👍

PR #177, opened 2020-06-27
Made NSwag output filename lowercase

MJM1989:
This fixes an issue (#176) with resolving the Angular client after initializing the template on a Linux machine (which has case-sensitivity for filepaths).


jasontaylordev:
Thanks 👍

PR #181, opened 2020-07-05
Private setter on collections in domain entity

DynamicCafeSolutions:
Implement the change made during the GOTO 2019 talk here:
https://youtu.be/dK4Yb6-LxAk?t=800

Private setters on collections prevent collection overriding.


jasontaylordev:
Thanks 👍

PR #184, opened 2020-07-09
Updating all .net packages and fix Fluent Validation breaking changes

mahmar:
Updating all the packages
Fixing Fluent Validation breaking changes as they remove n[on-generic](https://docs.fluentvalidation.net/en/latest/upgrading-to-9.html#removal-of-non-generic-validationcontext) `ValidationContext`.

Please review my changes and let me know if required any further improvement.
Thanks.


jasontaylordev:
Thank you!

PR #197, opened 2020-07-23
Rename ValidationBehavior to ValidationBehaviour

MHarrison22:
Renamed class to match non-american version of the word behavior for consistency.


jasontaylordev:
Thank you 👍

PR #210, opened 2020-08-09
Added rider cache/options directory to gitignore

MCCshreyas:
1. Added rider cache/options directory to gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,9 @@ Obj/
 # Visual Studio 2015 cache/options directory
 .vs/
 
+# Jetbrains Rider cache/options directory
+.idea/
+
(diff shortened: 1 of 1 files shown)


Leanwit:
Only to add information about this, there is another way to ignore this kind of files added by IDE (especially to avoid ignoring it in every project)
https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

MCCshreyas:
Very Thanks @Leanwit . I wasn't aware of a global `.gitignore` file. Surely, I will add this to my local machine!

jasontaylordev:
Thanks @MCCshreyas 😊

Issue #199, opened 2020-07-28
[Feature Request] Domain Events


scottkuhl:
We have added domain events to our application which is based on this solution template.  

Our domain events pattern provides for firing events as part of SaveChanges() in the database context that can participate in the same transaction.  

Alternatively, you can fire a domain event at any time.  For example, this provides for a case where you need to send an email message after the transaction is complete.   You would not want that message to go out during an open transaction and then have the transaction fail.

Is there interest in me creating a PR to integrate this into the template?

rajasingh012:
I think it should be nice to include domain events in this template

GFoley83:
Raising a domain event inside Save Changes is a bit limiting. I think a better strategy for domain events would be to add them as a post processor behaviour pipeline, as by then you know all logic in and around your handler has executed successfully.

rajasingh012:
Thanks @GFoley83 , So it seems classes are always anemic, we can add logic in post processor behavior pipeline or mediatr Notification. Is that recommended practice?

GFoley83:
Well what I've done in the past is both: a mediatr notification sent from a post processor behaviour pipeline. You can then have any number of classes in infrastructure listening to those events 
e.g. to add the event to a service bus or message queue, send an email etc.

scottkuhl:
I'll put together a brach soon to show what we have.

I agree that the event firing in Save Changes has limits which is why we have both methods. But the Save Changes fire also provides for an easy way to keep events in the same transaction when needed.

Im looking forward to seeing what enhancements the community has based on these conversations.

GFoley83:
>the Save Changes fire also provides for an easy way to keep events in the same transaction when needed.

Another strategy I've used in the past is to not expose `SaveChanges` to handlers at all (i.e. dont include a `SaveChanges` method on whatever interface you're injecting into your handlers for your `DbContext`) and instead call `SaveChanges` as part of a pipeline behavior e.g. as part of a Db transaction behavior pipeline. 

This can help insure that `SaveChanges` is only called when there are actually Db changes and also that it isn't being called multiple times within the same transaction. 

Of course, implementing something like this won't work for every scenario e.g. auto generated table Ids that get returned as part of the request response.

scottkuhl:
[...]

PR #217, opened 2020-08-17
[Feature] Domain Events

scottkuhl:
This PR adds Domain Events to the solution template.  Domain events implement side effects of changes across multiple aggregates, entities or services and provide very loose coupling.  Let's look at an example:

When a user places an order with an online store an email is sent back to the user confirming the purchase.  This could be done in a workflow in a single command where the data is saved to the database and then an email is sent.  Or you can handle the order placed as a command that throws an order placed event.  Then a separate event handler can send the email.  If in the future a new action is identified such as also sending an email to the fulfillment center, it can be added without changing the previous command.

There are two separate ways to fire domain events.

1. **During Save Changes.** This is based off Jimmy Bogart's [A Better Domain Events Pattern](https://lostechies.com/jimmybogard/2014/05/13/a-better-domain-events-pattern/). This is a simple and straightforward way to attach events to the domain model that are fired when the object is persisted to the database. Microsoft also has documentation on domain events with a similar implementation in the microservices architecture book in the section [Domain events: design and implementation](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/domain-events-design-implementation). An implementation and example of this pattern is included in the PR.

2. **Independent of Persistence.** This one is based on Steve Smith's [Immediate Domain Event Salvation with MediatR](https://ardalis.com/immediate-domain-event-salvation-with-mediatr/). It takes a different approach using a static method that lets you fire events at any time.  This option is **not** included in this PR.  If we want to add this it can be done with a static domain event class that has a Raise method on it.  I'm not sure if two different domain event handlers should be included in the template.

**Example Included**

The To Do Item Created Event is will be thrown by the create command.

```c#
    public class TodoItemCreatedEvent : DomainEvent
    {
        public TodoItemCreatedEvent(TodoItem item)
        {
            Item = item;
        }

        public TodoItem Item { get; }
    }
```

The To Do Item Completed Event Handle will catch the event.

```c#
    public class TodoItemCompletedEventHandler : INotificationHandler<DomainEventNotification<TodoItemCompletedEvent>>
    {
        private readonly ILogger<TodoItemCompletedEventHandler> _logger;

        public TodoItemCompletedEventHandler(ILogger<TodoItemCompletedEventHandler> logger)
        {
            _logger = logger;
        }

        public Task Handle(DomainEventNotification<TodoItemCompletedEvent> notification, CancellationToken cancellationToken)
        {
            var domainEvent = notification.DomainEvent;
            _logger.LogInformation("Domain event - {domainEvent} handler called", domainEvent.GetType().Name);

            return Task.CompletedTask;
        }
    }
````

The flow will work like this:

1. The Create To Do Command adds the event:

```c#
    ...
     entity.DomainEvents.Add(new TodoItemCreatedEvent(entity));
    ...
```

2. The event is fired during the Save Changes:

```c#
await DispatchEvents(cancellationToken);
```

3. The To Do Item Completed Handler is executed because it is specifically design to listen to that event:

```c#
INotificationHandler<DomainEventNotification<TodoItemCompletedEvent>>
```

### DIFF
--- a/src/Application/Common/Interfaces/IDomainEventService.cs
+++ b/src/Application/Common/Interfaces/IDomainEventService.cs
@@ -1,0 +1,10 @@
+﻿using CleanArchitecture.Domain.Common;
+using System.Threading.Tasks;
+
+namespace CleanArchitecture.Application.Common.Interfaces
+{
+    public interface IDomainEventService
+    {
+        Task Publish(DomainEvent domainEvent);
+    }
+}
--- a/src/Application/Common/Models/DomainEventNotification.cs
+++ b/src/Application/Common/Models/DomainEventNotification.cs
@@ -1,0 +1,15 @@
+﻿using CleanArchitecture.Domain.Common;
+using MediatR;
+
+namespace CleanArchitecture.Application.Common.Models
+{
+    public class DomainEventNotification<TDomainEvent> : INotification where TDomainEvent : DomainEvent
+    {
+        public DomainEventNotification(TDomainEvent domainEvent)
+        {
+            DomainEvent = domainEvent;
+        }
+
+        public TDomainEvent DomainEvent { get; }
+    }
+}
--- a/src/Application/TodoItems/Commands/CreateTodoItem/CreateTodoItemCommand.cs
+++ b/src/Application/TodoItems/Commands/CreateTodoItem/CreateTodoItemCommand.cs
@@ -1,5 +1,6 @@
 ﻿using CleanArchitecture.Application.Common.Interfaces;
 using CleanArchitecture.Domain.Entities;
+using CleanArchitecture.Domain.Events;
 using MediatR;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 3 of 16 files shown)


scottkuhl:
This PR is in response to [[Feature Request] Domain Events #199](https://github.com/jasontaylordev/CleanArchitecture/issues/199)

MichelJansson (commented):
Nice to add domain events!
There seem to be some changes here and there not related to this feature though. For instance [src/Domain/Common/AuditableEntity.cs](
https://github.com/jasontaylordev/CleanArchitecture/pull/217/files#diff-83e6c10128778e00757a50b6259eddf4) that is just reordering som properties for no apparent reason.

Maybe revert those changes to only include changes for this feature?

jasontaylordev:
Awesome stuff, thanks again @scottkuhl and @mccshreyas! 👍

agausachs:
QUESTION: Hi @scottkuhl, sorry to bring back such an old topic, I'm just trying to learn 😊 : What's the point of using DomainEventNotification instead of implementing INotification directly in DomainEvent? I can't see what it's the benefit. 

Thank you!

scottkuhl:
@agausachs, The original implementation of this did in fact implement INotification on the Domain Event.  It was suggested in the feature request issue that we remove the dependency on MediatR in the domain project as it was not there before.  The full conversion which may answer your question can be found in the feature request issue: #199

agausachs:
After reading the thread #199, it totally make sense. Thanks!

PR #220, opened 2020-08-27
added generic exception  to handle some non validation errors

abdulateef:


jasontaylordev:
Thanks for your contribution @abdulateef. Since `ErrorCode` is not used, what about switching to `System.ApplicationException` instead of a new `GenericException`? An `ApplicationExcepiton` thrown in the `Application` layer seems like a good approach. Thoughts?

Also, how will you use this exception? What exceptional scenarios will be handled this way? Do you have an example?

abdulateef:
> Thanks for your contribution @abdulateef. Since `ErrorCode` is not used, what about switching to `System.ApplicationException` instead of a new `GenericException`? An `ApplicationExcepiton` thrown in the `Application` layer seems like a good approach. Thoughts?
> 
> Also, how will you use this exception? What exceptional scenarios will be handled this way? Do you have an example?



> Thanks for your contribution @abdulateef. Since `ErrorCode` is not used, what about switching to `System.ApplicationException` instead of a new `GenericException`? An `ApplicationExcepiton` thrown in the `Application` layer seems like a good approach. Thoughts?
> 
> Also, how will you use this exception? What exceptional scenarios will be handled this way? Do you have an example?

thanks @jasontaylordev 
the idea is to be able to return certain information from the application that are not validation exceptiond
`
var category = Context.Find<Category>(productDto.CategoryId);
            if (category == null )
            {
                throw new GenericException("category does not exist");
            }
`
the code  above is what i intended.

abdulateef:
thanks @jasontaylordev 
the idea is to be able to return certain information from the application that are not validation exceptions
`
          var category = Context.Find<Category>(productDto.CategoryId);
            if (category == null )
            {
                throw new GenericException("category does not exist");
            }
`
the code  above is what i intended.

deceptiveSimplicity:
Hello @jasontaylordev 
I tend to avoid ApplicationException because of a remark in the documentation.
[https://docs.microsoft.com/en-us/dotnet/api/system.applicationexception?view=netcore-3.1](url)
> You should derive custom exceptions from the Exception class rather than the ApplicationException class. You should not throw an ApplicationException exception in your code, and you should not catch an ApplicationException exception unless you intend to re-throw the original exception.

jasontaylordev:
Thanks @abdulateef and @deceptiveSimplicity. I understand the concept behind the GenericException, it's one size fits many types approach. I can see now that ApplicationException shouldn't be used, although it has the perfect name for throwing exceptions from the Application project. ☹

The example you've provided is for a not found exception. We already have a NotFoundException to handle that circumstance. In general, I'd prefer to use a named exception such as NotFound, Validation, etc, rather than a Generic Exception. 

I'd consider creating a CoreException (since Domain and Application represent the Core of the system). This would be defined in Domain and could be thrown in Application or Domain. It would represent a complete failure.

I think for the time being I'll leave GenericException out, but let me know if I have missed something. 

Regardless, appreciate your contribution and the discussion.

Happy Coding! 😊

PR #224, opened 2020-08-29
ValidateModelFilterAttribute added.

iayti:
ValidateModelFilterAttribute checks Request Object Json convertion and response BadRequest(error.message)

Sample Response: id : "The JSON value could not be converted to System.Int32. Path: $.id | LineNumber: 1 | BytePositionInLine: 13."

Thank you for your interest. Have a nice coding 👍

### DIFF
--- a/src/WebUI/Filters/ValidateModelFilterAttribute.cs
+++ b/src/WebUI/Filters/ValidateModelFilterAttribute.cs
@@ -1,0 +1,16 @@
+﻿namespace CleanArchitecture.WebUI.Filters
+{
+    using Microsoft.AspNetCore.Mvc;
+    using Microsoft.AspNetCore.Mvc.Filters;
+
+    public class ValidateModelFilterAttribute : ActionFilterAttribute
+    {
+        public override void OnActionExecuting(ActionExecutingContext context)
+        {
+            if (context.ModelState.IsValid)
+                return;
+
+            context.Result = new BadRequestObjectResult(context.ModelState);
+        }
+    }
+}
--- a/src/WebUI/Startup.cs
+++ b/src/WebUI/Startup.cs
@@ -41,8 +41,10 @@ namespace CleanArchitecture.WebUI
                 .AddDbContextCheck<ApplicationDbContext>();
 
             services.AddControllersWithViews(options =>
-                options.Filters.Add(new ApiExceptionFilterAttribute()))
-                    .AddFluentValidation();
+            {
+                options.Filters.Add(new ApiExceptionFilterAttribute());
+                options.Filters.Add(new ValidateModelFilterAttribute());
+            }).AddFluentValidation();
 
             services.AddRazorPages();
 


MCCshreyas:
Hi there @iayti, we already have FluentValidation in place to validate the incoming Commands & Queries to our API. I don't think so adding this extra layer will have any major benefit.

iayti:
> Hi there @iayti, we already have FluentValidation in place to validate the incoming Commands & Queries to our API. I don't think so adding this extra layer will have any major benefit. May I know any specific reason for adding this! 😃

if the request object json format is true, "FluentValidation" works very well. If the format is wrong, response will be 500 internal error.

I don't think this response is accurate and revealing enough. Because the real error is client originated but we show it as server error.

Sample Post: https://localhost:5001/api/ToDoItems

{
  "listId": "1",
  "title": "Test"
}

Current Sample Response:
{
    "type": "https://tools.ietf.org/html/rfc7231#section-6.6.1",
    "title": "An error occurred while processing your request.",
    "status": 500
}

My Solution Sample Response: 400 Bad Request
{
    "$.listId": [
        "The JSON value could not be converted to System.Int32. Path: $.listId | LineNumber: 1 | BytePositionInLine: 15."
    ]
}

If FluentValidation can solve this problem, we do not need to add this layer as @MCCshreyas  said 😃


Screen Shot 2020-08-30 at 20 32 47

jasontaylordev:
Hi @iayti - thanks for identifying this issue. To resolve I've taken a different approach - see https://github.com/jasontaylordev/CleanArchitecture/commit/f8e12018011051d3078ccc71e260aa9604c1f275. This way it's just using the `ApiExceptionFilterAttribute` and the response is returned as a problem details response. For example:

```json
{
  "type": "https://tools.ietf.org/html/rfc7231#section-6.5.1",
  "title": "One or more validation errors occurred.",
  "status": 400,
  "errors": {
    "$.listId": [
      "The JSON value could not be converted to System.Int32. Path: $.listId | LineNumber: 0 | BytePositionInLine: 13."
    ]
  }
}
```

Let me know what you think. Appreciate your contribution. 👍

iayti:
> Hi @iayti - thanks for identifying this issue. To resolve I've taken a different approach - see [f8e1201](https://github.com/jasontaylordev/CleanArchitecture/commit/f8e12018011051d3078ccc71e260aa9604c1f275). This way it's just using the `ApiExceptionFilterAttribute` and the response is returned as a problem details response. For example:
> 
> ```json
> {
>   "type": "https://tools.ietf.org/html/rfc7231#section-6.5.1",
>   "title": "One or more validation errors occurred.",
>   "status": 400,
>   "errors": {
>     "$.listId": [
>       "The JSON value could not be converted to System.Int32. Path: $.listId | LineNumber: 0 | BytePositionInLine: 13."
>     ]
>   }
> }
> ```
> 
> Let me know what you think. Appreciate your contribution. 👍

Thank you for the answer, this solution is much more beautiful 😊

PR #232, opened 2020-09-09
Mapping test should handle entities without parameterless constructor

mmaterowski:
It is oftentimes the case, that You don't allow entities with parameterless constructor in Your project to prevent developers for creating invalid entities. I found mapping test very elegant and I see a benefit for other developers, which might wanna use them in their projects.

GetUninitialized method creates an instance of an object and set all fields to default value, but do not call any of constructors. So mapping tests are still valid, but we skip any constructor-related issues if we cannot or don't have access to parameterless constructor


jasontaylordev:
Thanks, I've just changed it to check for a parameterless constructor rather than using a try ... catch.

PR #242, opened 2020-09-20
Fix DispatchEvents() that caused an infinite loop of events

nikibreg:
In case there's a domain event handler that's also triggering another domain event, it will enter an infinite loop.

To fix this, I am checking if the domain event has already been fired after its initialization.

### DIFF
--- a/src/Domain/Common/DomainEvent.cs
+++ b/src/Domain/Common/DomainEvent.cs
@@ -14,7 +14,7 @@ namespace CleanArchitecture.Domain.Common
         {
             DateOccurred = DateTimeOffset.UtcNow;
         }
-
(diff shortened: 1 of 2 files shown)


GFoley83:
I feel this issue is a side effect of having the domain events dispatched from the DbContext/Infrastructure layer in the first place.

I think a better design would be to dispatch events manually from the Command handler, e.g. after the `SaveChanges` call 
or
from a behavior pipeline that runs after the command has executed successfully.

jasontaylordev:
Thanks @nikibreg - I'll merge this now. 👍 

Thanks @GFoley83 - I'll add your suggestion to the list. 😀

PR #243, opened 2020-09-29
Provide enum values

sMustafov:
It is very common to discover that you need to add values to an enum after you have already shipped it. There is a potential application compatibility problem when the newly added value is returned from an existing API, because poorly written applications might not handle the new value correctly.


Leanwit (approved):

jasontaylordev:
Thank you 😊

PR #244, opened 2020-10-03
Done with general code improvements and added some unit tests

MCCshreyas:
1. Removed unwanted `usings` .
2. Done some changes to Domain models for immutability. 
3. Used some `async` overload in `async` handlers. 
4. Added a `Guid` field for Event abstraction to recognize each domain event. 
5. Used pattern matching. 
5. Added some more domain unit tests.


src/Domain/Common/ValueObject.cs
```diff
@@ -31,8 +31,8 @@ public override bool Equals(object obj)
             }
 
             var other = (ValueObject)obj;
-            var thisValues = GetAtomicValues().GetEnumerator();
-            var otherValues = other.GetAtomicValues().GetEnumerator();
+            using var thisValues = GetAtomicValues().GetEnumerator();
+            using var otherValues = other.GetAtomicValues().GetEnumerator();
```
Leanwit, line 35: What is the difference between `var thisValues` vs `using var thisValues` in this context?

Leanwit (approved):

src/Domain/Common/ValueObject.cs
MCCshreyas: it's under `using` code block, so the objects will get disposed at the end of the parent code block.

src/Application/TodoItems/Commands/CreateTodoItem/CreateTodoItemCommand.cs
```diff
@@ -25,16 +25,11 @@ public CreateTodoItemCommandHandler(IApplicationDbContext context)
...
-                ListId = request.ListId,
-                Title = request.Title,
-                Done = false
-            };
+            var entity = new TodoItem(request.ListId, request.Title);
 
             entity.DomainEvents.Add(new TodoItemCreatedEvent(entity));
 
-            _context.TodoItems.Add(entity);
+            await _context.TodoItems.AddAsync(entity, cancellationToken);
```
hummer007, line 32: The AddAsync method should be used only in certain cases, you can read about it on the Microsoft website

jasontaylordev:
Hi @MCCshreyas - thanks for your efforts and patience. 🙏

I can see that this pull request includes many changes - such as implementing a rich domain model with immutability, unit testing, general refactoring, and so on. It would be much better if the request could be split into numerous small requests. 

Please note that:

1. I am not looking to demonstrate a rich domain model with immutability
2. `DbSet.AddAsync` is only for special cases, see https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbcontext.addasync?view=efcore-5.0
3. The AdAccount value object has been removed and replaced with a Colour value object

I try to keep the template as simple as possible, and so I have to pick and choose which concepts to include. I know this makes it difficult to know where contributions can be made, my apologies. Feel free to reach out to me if you want to brainstorm some ideas. 

In the meantime, for any of the other concepts in this pull request, please create a smaller pull request for review.

PR #251, opened 2020-10-18
added a few unit tests. Used MockQueryable to directly mock the DbSets

Alexei000:
This adds a few unit tests to act as an example for those using the template. This is particularly important to show that it is easy to unit test despite not using a generic repository. This is possible because DbSet mocking is very simply due to [MockQueryable](https://github.com/romantitov/MockQueryable).

### DIFF
--- a/src/Application/Common/Extensions/EnumerableExtensions.cs
+++ b/src/Application/Common/Extensions/EnumerableExtensions.cs
@@ -1,0 +1,15 @@
+﻿using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace CleanArchitecture.Application.Common.Extensions
+{
+    public static class EnumerableExtensions
+    {
+        public static IEnumerable<T> DeepClone<T>(this IEnumerable<T> list)
+        {
+            string serializedData = JsonConvert.SerializeObject(list);
+            var newList = JsonConvert.DeserializeObject<IEnumerable<T>>(serializedData);
+            return newList;
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


src/Application/Common/Extensions/EnumerableExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+    public static class EnumerableExtensions
+    {
+        public static IEnumerable<T> DeepClone<T>(this IEnumerable<T> list)
+        {
+            string serializedData = JsonConvert.SerializeObject(list);
+            var newList = JsonConvert.DeserializeObject<IEnumerable<T>>(serializedData);
+            return newList;
+        }
+    }
+}
```
Leanwit, line 15: [Doubt] Is the application project the correct place for this extension as you are only using the code for testing purposes? Is it better to move this code to Application if someday we will use the DeepClone method there? I had this thought and I would like to know other points of view.

Alexei000: Hello! 

Yes, I have introduced it there because it is very generic and can be used in the future to easily deeply clone any IEnumerable.

Of course, I can keep it in the test project only. I am not sure where these utility functions should stay in this solution.

Alexei000: I have moved EnumerableExtensions into Application.UnitTests project, so that all the changes are now done only to this project.

jasontaylordev:
Hi @Alexei000 - thanks for this contribution. 

I prefer to use an InMemory provider such as EF Core or SQLite InMemory for unit tests. These have been built for testing purposes. While these providers don't behave like a real database, that's fine for the purposes of these tests. We're not testing the database or EF Core, just some command or query that we've written. That being the case we just need a provider to behave well enough to support unit testing. They do this fairly well.

This is a talk I did on it a while ago - https://www.youtube.com/watch?v=ddrR440JtiA.

PR #259, opened 2020-11-07
Update LoggingBehaviour.cs

nikibreg:
Fix an incorrect check that redundantly reassigned the user ID


src/Application/Common/Behaviours/LoggingBehaviour.cs
```diff
@@ -25,7 +25,7 @@ public async Task Process(TRequest request, CancellationToken cancellationToken)
             var userId = _currentUserService.UserId ?? string.Empty;
             string userName = string.Empty;
 
-            if (!string.IsNullOrEmpty(userId))
+            if (string.IsNullOrEmpty(userId))
```
Leanwit, line 28: In my opinion, this is not a bug because you need an UserId to find the username. In this case, you are executing the GetUserNameAsync method sending a null value as a parameter.

PR #271, opened 2020-11-30
.Net 5 Fixed Build issue caused by  Nswag MSBuild  runtime variable configuration 

hnicolus:
The MSBuild  configurations for NSwagg command line was still trying to use Net core 3.1 runtime.
It was causing my builds to fail. The fix was to change the  $NSwagExe_31 variable  to $NSwagExe_50 to 
For more information you can visit [NSwag.MSBuild](https://github.com/RicoSuter/NSwag/wiki/NSwag.MSBuild) 🙂


jasontaylordev:
Thank you 😊

PR #279, opened 2020-12-14
use routing module for main app module

kingjordan:
Hello,

this is my first public pull request so forgive me if I get something wrong.

I moved the routing paths from the app.module file into a separate app-routing.module.ts file as that is likely what every developer will do once the app has more than a couple routes.  Its also best practice according to[ angular.io. ](https://angular.io/guide/router) 
One final benefit is if you try to add something like nativescript, it can better discover the existing routes when creating the new nativescript code sharing project.

This is a preference change and does not change app functionality so integrate it if you like

Cheers


jasontaylordev:
Thanks 😊

PR #281, opened 2020-12-15
Update angular 11

kingjordan:
Hello,

I have upgraded the Angular App to 11 which is the most recent.  I also cleaned up some small code as a result of the upgrade. 
ie: changing async to waitForAsync in some angular .spec.ts files

Updated the Readme file with links to the technologies used


kingjordan:
Hey Jason, I'm going to drop this, I jumped the gun and there are newer versions of angular compared to this.

jasontaylordev:
No worries, thanks anyway.

PR #284, opened 2020-12-16
Dispatch domain events from a behaviour

nikibreg:
As suggested by @GFoley83 in #242, I removed domain event dispatching from the db context and instead created a pipeline behaviour that runs after requests to do it

### DIFF
--- a/src/Application/Common/Behaviours/DomainEventPublishingBehavior.cs
+++ b/src/Application/Common/Behaviours/DomainEventPublishingBehavior.cs
@@ -1,0 +1,46 @@
+﻿using CleanArchitecture.Application.Common.Interfaces;
+using CleanArchitecture.Domain.Common;
+using MediatR.Pipeline;
+using Microsoft.EntityFrameworkCore.ChangeTracking;
+using Microsoft.Extensions.Logging;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace CleanArchitecture.Application.Common.Behaviours
+{
+    public class DomainEventPublishingBehaviour<TRequest, TResponse> : IRequestPostProcessor<TRequest, TResponse>
+    {
+        private readonly ILogger _logger;
+        private readonly IApplicationDbContext _dbContext;
+        private readonly IDomainEventService _domainEventService;
+
+        public DomainEventPublishingBehaviour(
+            ILogger<TRequest> logger,
+            IApplicationDbContext dbContext,
+            IDomainEventService domainEventService)
+        {
+            _logger = logger;
+            _dbContext = dbContext;
+            _domainEventService = domainEventService;
+        }
+
+        public async Task Process(TRequest request, TResponse response, CancellationToken cancellationToken)
+        {
+            while (true)
+            {
+                var domainEventEntity = _dbContext.ChangeTracker.Entries<IHasDomainEvent>()
+                                                     .Select(x => x.Entity.DomainEvents)
+                                                     .SelectMany(x => x)
+                                                     .Where(domainEvent => !domainEvent.IsPublished)
+                                                     .FirstOrDefault();
+                if (domainEventEntity == null) break;
+
+                domainEventEntity.IsPublished = true;
+                await _domainEventService.Publish(domainEventEntity);
+                
+                _logger.LogInformation("Published event: {Name}", domainEventEntity.GetType().Name);
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


io383279 (approved):

neman:
What about integration events?
What if you want transactional save of events to make sure they are preserved together when saving domain object.
Having this in behavior it is impossible to use outbox pattern.

https://microservices.io/patterns/data/transactional-outbox.html

I do not see why we should move this from DbContext when there are many examples from experts in this field

https://docs.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/domain-events-design-implementation
https://lostechies.com/jimmybogard/2014/05/13/a-better-domain-events-pattern/
https://devblogs.microsoft.com/cesardelatorre/domain-events-vs-integration-events-in-domain-driven-design-and-microservices-architectures/

Why do we need to reinvent the wheel or put everything in behaviors? What would you do if you do not use MediatR? Put this in middleware or use decorator pattern. Dispatching event should be on the lowest level after transaction is committed.

MCCshreyas:
Pipelines are used for cross-cutting concerns. Injecting DBContext there, won't look good from my point of view.

jtanios:
@neman 
> I do not see why we should move this from DbContext when there are many examples from experts in this field

I think the main reason to move it is an architectural one. The current implementation of domain event dispatching live in the **Infrastructure** layer, while Domain Events belong to the **Application** layer. Currently, we're not able to test the Application layer unless this event dispatching behavior is also implemented in the InMemoryDB. The Application layer should be able to function independently of any implementation of DBContext.

pcarvsilva:
I agree with @neman in the sense that it should be in infrastructure, I was thinking into inserting [Greg Young's event store](https://www.eventstore.com/) to persist domain events. Maybe the proper solution would be to persist and infrastructure and then send it back to the domain. In this way you can make Materialized Views with Mongo reading directly from the event store or adding stuff into another infrastructures without event noticing what happened into the Application

neozhu:
There is a slight problem when ApplicationDbContext DI ServiceLifetime is Transient.  the ApplicationDbContext  are two different instances

neman:
> There is a slight problem when ApplicationDbContext DI ServiceLifetime is Transient. the ApplicationDbContext are two different instances

Why would you use transient DbContext for scoped service? DbContext is UnitOfWork, its main purpose is to keep track of everything you change https://martinfowler.com/eaaCatalog/unitOfWork.html

If you do not want, you are doing something wrong. Give me an example of your scenario?
I have used both scoped and transient DbContext, depending on context, and had no problem. 
I used transient for some workers.

neozhu:
@neman 
I know, I work it in blazor server project.  I have been wondering how to use ef core in blazor Server project .

src/Application/Common/Behaviours/DomainEventPublishingBehavior.cs
```diff
@@ -0,0 +1,46 @@
...
+                                                     .Where(domainEvent => !domainEvent.IsPublished)
+                                                     .FirstOrDefault();
+                if (domainEventEntity == null) break;
+
+                domainEventEntity.IsPublished = true;
+                await _domainEventService.Publish(domainEventEntity);
+                
+                _logger.LogInformation("Published event: {Name}", domainEventEntity.GetType().Name);
+            }
+        }
```
Bardin08, line 44: Using an infinity loop looks like not the best solution. Just my preference not to use `while-true loops` if it's possible to avoid them.

For example, this code is entirely like yours but without infinity looks.

```cs
var domainEventEntities = _dbContext.ChangeTracker.Entries<IHasDomainEvent>()
                                  .Select(x => x.Entity.DomainEvents)
                                  .SelectMany(x => x)
                                  .Where(domainEvent => !domainEvent.IsPublished);
                    
if (domainEventEntities is null)
	return;
              
foreach(var @event in domainEventEntities)
{
	domainEventEntity.IsPublished = true;
    await _domainEventService.Publish(domainEventEntity);

	_logger.LogInformation("Published event: {Name}", domainEventEntity.GetType().Name);
}
```

Bardin08 (approved):
@nikibreg that's a really good improvement that helps to keep the template's clarity higher

jasontaylordev:
Thanks for submitting this PR and for the discussion around the implementation. Apologies for taking so long to review and decide a course of action...
For this solution the goal has always been the simplest approach to enterprise architecture. In keeping with that theme, and creating a template that is useful for many solutions, I've decided to keep event dispatching within infrastructure.
I've recently pushed through some changes to clean-up the current implementation, please feel free to review and provide feedback. It's clearly not a one size fits all approach, but a good starting point regardless. Happy coding! 😀

PR #314, opened 2021-02-28
Use `record` instead of `class` for ValueObject

jtanios:
This takes advantage of the new `record` type that encapsulates value object equality.

Benefits
* No code in `ValueObject`
* Classes inheriting `ValueObject` (i.e. `Colour`) don't need to implement `IEnumerable<object> GetEqualityComponents()`

### DIFF
--- a/src/Domain/Common/ValueObject.cs
+++ b/src/Domain/Common/ValueObject.cs
@@ -1,44 +1,6 @@
-﻿using System.Collections.Generic;
-using System.Linq;
-
-namespace CleanArchitecture.Domain.Common
+﻿namespace CleanArchitecture.Domain.Common
 {
-    // Learn more: https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/microservice-ddd-cqrs-patterns/implement-value-objects
-    public abstract class ValueObject
+    public abstract record ValueObject
     {
-        protected static bool EqualOperator(ValueObject left, ValueObject right)
-        {
-            if (left is null ^ right is null)
-            {
-                return false;
-            }
-
-            return left?.Equals(right) != false;
-        }
-
-        protected static bool NotEqualOperator(ValueObject left, ValueObject right)
-        {
-            return !(EqualOperator(left, right));
-        }
-
-        protected abstract IEnumerable<object> GetEqualityComponents();
-
-        public override bool Equals(object obj)
-        {
-            if (obj == null || obj.GetType() != GetType())
-            {
-                return false;
-            }
-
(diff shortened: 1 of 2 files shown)


MCCshreyas:
Am sure we can use records for CQRS command and query classes as well! 😅😅

jtanios:
> Am sure we can use records for CQRS command and query classes as well!

Interesting. What would be the advantage of doing that?

jtanios:
> Am sure we can use records for CQRS command and query classes as well!

Interesting. What would be the advantage of doing that?

jtanios:
> Am sure we can use records for CQRS command and query classes as well!

Interesting. What would be the advantage of doing that?

jtanios:
Any interest in merging this?

achmstein:
Please read this article https://enterprisecraftsmanship.com/posts/csharp-records-value-objects/

jamesSampica:
> 
> 
> > Am sure we can use records for CQRS command and query classes as well!
> 
> Interesting. What would be the advantage of doing that?

Easier to write and test. For example...

    public class UpdateTodoItemCommand : IRequest
    {
        public int Id { get; set; }

        public bool Done { get; set; }
    }

becomes `public record UpdateTodoItemCommand(int Id, string Title, bool Done) : IRequest`

    var itemId = await SendAsync(new CreateTodoItemCommand
    {
        ListId = listId,
        Title = "New Item"
    });

becomes `await SendAsync(new CreateTodoItemCommand(listId, "New Item"))`

I don't see any WebUI tests but verifying controllers invoke certain actions through mocking frameworks is also easier

using `class`

//arrange
var query = new DoThingCommand { SomeProperty = "Some String" };
//act
controller.DoThing(query.SomeProperty);
//assert
mediatorMock.Verify(mock => mock.Send(It.Is<DoThingCommand>(c => c.SomeProperty == "Some String"));

using `record`

//arrange
var query = new DoThingCommand("Some String");
//act
controller.DoThing(query.SomeProperty);
//assert
mediatorMock.Verify(mock => mock.Send(query));

jasontaylordev:
> Please read this article https://enterprisecraftsmanship.com/posts/csharp-records-value-objects/

This is a great post. In my mind, record types offer ultimate simplicity whereas value objects offer complete control. I'd like to demonstrate both in the template.  So for the time being I think a good approach will be to keep value objects at a domain level and switch to record types for commands and queries (see #348).

Thanks, @jtanios - it's not precisely the change you set out to make, but the contribution is still appreciated. ❤

PR #319, opened 2021-03-07
Feature/data protection

naster01:
### Context
The main purpose of this PR is to protect data leaks from one user to an other by automatically adding query filters to entities. The filter can be removed with `_context.TodoItems.IgnoreQueryFilters()` to query records belonging to all users.

Example of Sql generated:
```sql
SELECT [t].[Id], [t].[Colour_Code], [t].[Title], [t1].[Done], [t1].[Id], [t1].[ListId], [t1].[Note], [t1].[Priority], [t1].[Title]
FROM [TodoLists] AS [t]
LEFT JOIN (
    SELECT [t0].[Done], [t0].[Id], [t0].[ListId], [t0].[Note], CAST([t0].[Priority] AS int) AS [Priority], [t0].[Title]
    FROM [TodoItems] AS [t0]
    WHERE [t0].[CreatedByUserId] = @__ef_filter__UserId_1
) AS [t1] ON [t].[Id] = [t1].[ListId]
WHERE [t].[CreatedByUserId] = @__ef_filter__UserId_0
ORDER BY [t].[Title], [t].[Id], [t1].[Id]
```

### What has been modified

- `ApplicationUser` was moved to the Domain layer as the base class `IdentityUser` is only a POCO and all properties are virtual.
- `AuditableEntity.CreatedBy` and `AuditableEntity.LastModifiedBy` were modified to use Ef Core naming convention.
- Foreing keys and naviguation properties were added to `AuditableEntity`.
- To reflect changes to the database, a migration was added.
- Global query filters are automatically added to entities requiring protection over data. `ApplicationDbContext.OnModelCreating`.


MCCshreyas:
We want our Domain project to be not dependent on any third party or any other infrastructure thing! In this case, your domain project is now dependent upon Microsoft identity which is *not* good thing!

naster01:
@MCCshreyas I've made some changes to remove dependencies to the domain layer and updated the PR description. What do you think?

naster01:
References to `Microsoft.Extensions.Identity.Stores` would be ok in Application layer. It only define adstractions.
https://github.com/dotnet/aspnetcore/tree/main/src/Identity/Extensions.Stores/src

src/Infrastructure/Identity/IdentityService.cs
```diff
@@ -74,7 +85,7 @@ public async Task<Result> DeleteUserAsync(string userId)
             return Result.Success();
         }
 
-        public async Task<Result> DeleteUserAsync(ApplicationUser user)
+        public async Task<Result> DeleteUserAsync(User user)
```
MrKoster, line 88: Shouldn't there be another user synchronization after deletion (same as after adding new user)? Something like:
```csharp
    public async Task<Result> DeleteUserAsync(User user)
    {
        var result = await _userManager.DeleteAsync(user);
        if (result.Succeeded)
        {
            // Sync identity user and application user
            var userToDelete = _context.ApplicationUsers.FirstOrDefault(x => x.Id == user.Id);
            if (userToDelete != null)
            {
                _context.ApplicationUsers.Remove(userToDelete);
                await _context.SaveChangesAsync(CancellationToken.None);
            }     
        }
        return result.ToApplicationResult();
    }
```

PR #335, opened 2021-03-28
Added token viewer component

matt-goldman:
This pull request adds a JWT viewer to the Angular app. This can be copied to the clipboard which can be useful for development (e.g. you can use the token in the Swagger UI page or in Postman).


jasontaylordev:
Thanks, @matt-goldman! ❤

PR #336, opened 2021-03-30
Update README.md

RazvanBd:
Update readme with `npm install`  instead of "run". Inspired by @kingjordan looking at his change on 5.0 branch


jasontaylordev:
Thanks, @RazvanBd. However, npm start is required as per @kingjordan's comment. Please check the current README which has both.

PR #346, opened 2021-04-16
generate nswag typescript code not only in debug build configuration

maxschaf:
Our Angular Frontend uses the generated TS Interfaces. If backend changes, this file gets regenerated, but only in Debug Mode. CI builds are run in release mode and after merging to main the merged TS files are potentially wrong and should be generated during release builds. 

This is a bug out of my perspective.

### DIFF
--- a/src/WebUI/WebUI.csproj
+++ b/src/WebUI/WebUI.csproj
@@ -1,85 +1,85 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
-  <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
-    <RootNamespace>CleanArchitecture.WebUI</RootNamespace>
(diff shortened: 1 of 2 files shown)


jasontaylordev:
> Our Angular Frontend uses the generated TS Interfaces. If backend changes, this file gets regenerated, but only in Debug Mode. CI builds are run in release mode and after merging to main the merged TS files are potentially wrong and should be generated during release builds.
> 
> This is a bug out of my perspective.

Thanks, @maxschaf - I can see that for some cases this would be an issue. However, this approach works for most cases and means that there is no need to run NSwag as part of the build pipeline. I also like seeing the generated changes to the spec/clients when I commit changes. 

I try to take the simplest approach in this template since it's easier to build on top of simplicity. Thanks anyway!

Issue #347, opened 2021-04-19
Use "record" for CQRS Queries and Commands


mus-h:
Use "record" for CQRS Queries and Commands, it will be simpler and cleaner and less prone to mistake.

**Additional context**
![Before and after](https://user-images.githubusercontent.com/30037608/115197941-349e7480-a0fa-11eb-8d01-3d5e0ce6f107.png)

PR #348, opened 2021-04-19
Replace all commands and queries to be records

misha130:
Based on the #347 and the discussion in #314 

This PR only applies to the commands and queries and does not change anything else.

The pros: 

* Shorter code
* Commands/Queries definitely fit the requirements to be records
* Can explicitly say what is required and what is optional

The cons:

* Readability is slightly lower since you can't read what is the property assigned directly by looking. (`new Command { Name = "Something" }` vs `new Command("Something")` )
* The user of this template might misunderstand that all commands/queries must be records which is not the case

### DIFF
--- a/src/Application/TodoItems/Commands/CreateTodoItem/CreateTodoItemCommand.cs
+++ b/src/Application/TodoItems/Commands/CreateTodoItem/CreateTodoItemCommand.cs
@@ -7,12 +7,7 @@ using System.Threading.Tasks;
 
 namespace CleanArchitecture.Application.TodoItems.Commands.CreateTodoItem
 {
-    public class CreateTodoItemCommand : IRequest<int>
-    {
-        public int ListId { get; set; }
-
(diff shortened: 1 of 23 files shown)


kingjordan (approved):
the built in immutability and equality checking of records is great for this

jasontaylordev:
Thank you. I think it would be good to take this approach: 

```csharp
    public record UpdateTodoItemDetailCommand : IRequest
    {
        public int Id { get; init; }

        public int ListId { get; init; }

        public PriorityLevel Priority { get; init; }

        public string Note { get; init; }
    }
```

Since we can use standard property syntax with records, readability will still be great. Because `init` is used instead of `set` the properties are still immutable.

misha130:
Thats fine. 
To be honest I was aiming for shorter code but I agree there is a benefit to your suggestion. 
I'll close this PR rather than rewriting it here

PR #374, opened 2021-06-23
Removed string typename and support for more interfaces

janosorcsik:
Hi @jasontaylordev !

This PR replaces the string type names with the interface type.
And it supports multiple interfaces.

### DIFF
--- a/src/Application/Common/Mappings/MappingProfile.cs
+++ b/src/Application/Common/Mappings/MappingProfile.cs
@@ -14,19 +14,40 @@ namespace CleanArchitecture.Application.Common.Mappings
 
         private void ApplyMappingsFromAssembly(Assembly assembly)
         {
-            var types = assembly.GetExportedTypes()
-                .Where(t => t.GetInterfaces().Any(i => 
-                    i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IMapFrom<>)))
-                .ToList();
+            var mapFromType = typeof(IMapFrom<>);
+            
+            var mappingMethodName = nameof(IMapFrom<object>.Mapping);
+
+            bool HasInterface(Type t) => t.IsGenericType && t.GetGenericTypeDefinition() == mapFromType;
+            
+            var types = assembly.GetExportedTypes().Where(t => t.GetInterfaces().Any(HasInterfaces)).ToList();
+            
+            var argumentTypes = new Type[] { typeof(Profile) };
 
             foreach (var type in types)
             {
                 var instance = Activator.CreateInstance(type);
-
-                var methodInfo = type.GetMethod("Mapping") 
-                    ?? type.GetInterface("IMapFrom`1").GetMethod("Mapping");
                 
-                methodInfo?.Invoke(instance, new object[] { this });
+                var methodInfo = type.GetMethod(mappingMethodName);
+
+                if (methodInfo != null)
+                {
+                    methodInfo.Invoke(instance, new object[] { this });
+                }
+                else
+                {
+                    var interfaces = type.GetInterfaces().Where(HasInterface).ToList();
+
+                    if (interfaces.Count > 0)
+                    {
+                        foreach (var @interface in interfaces)
+                        {
+                            var interfaceMethodInfo = @interface.GetMethod(mappingMethodName, argumentTypes);
+
+                            interfaceMethodInfo?.Invoke(instance, new object[] { this });
+                        }
+                    }
+                }
 
             }
         }


jasontaylordev:
I like it - thank you. I can think of a few (e.g. lookups), but what scenarios do you use multiple interfaces for?

janosorcsik:
I use an IMapTo interface for mapping DTOs to entities. Same as IMapFrom in this project.

For example, you have an edit form for an entity. When you call the server for data you got a DTO and you can use this DTO for saving the entity.
The GetEntityQuery returns the DTO and the SaveEntityCommand inherits from the DTO.

Eagleleader:
I'm interested,

Do you apply IMapTo into the DTO too and this is why you changed for multiples interfaces ?

neozhu:
Is it okay to change it?
```js
void Mapping(Profile profile) => profile.CreateMap(typeof(T), GetType(),MemberList.None).ReverseMap();
```
by the way, here's how to avoid this question what "The following type maps were found in multiple profiles:" in unit testing.
following code :
```js
public partial class DocumentDto : IMapFrom<Document>
    {
        public void Mapping(Profile profile)
        {
           // here, I need to define a special mapping (e.g)
            profile.CreateMap<Document, DocumentDto>()
                .ForMember(x => x.DocumentTypeName, s => s.MapFrom(y => y.DocumentType.Name));
            profile.CreateMap<DocumentDto, Document>()
                .ForMember(x => x.DomainEvents, y => y.Ignore())
                .ForMember(x => x.DocumentType, s => s.Ignore())
                .ForMember(x => x.Created, y => y.Ignore())
                .ForMember(x => x.CreatedBy, y => y.Ignore())
                .ForMember(x => x.LastModified, y => y.Ignore())
                .ForMember(x => x.LastModifiedBy, y => y.Ignore());
        }
        public int Id { get; set; }
        public string Description { get; set; }
        public bool IsPublic { get; set; } = false;
        public string URL { get; set; }
        public string CreatedBy { get; set; }
        public DateTime Created { get; set; }
        public int DocumentTypeId { get; set; }
        public string DocumentTypeName { get; set; }
    }
```
unit test output, In this case the “ Mapping(Profile profile)” method was executed twice.
```
ShouldHaveValidConfiguration
   Source: MappingTests.cs line 33
   Duration: 26 ms

  Message: 
AutoMapper.DuplicateTypeMapConfigurationException : The following type maps were found in multiple profiles:
CleanArchitecture.Razor.Domain.Entities.Document to CleanArchitecture.Razor.Application.Documents.DTOs.DocumentDto defined in profiles:
CleanArchitecture.Razor.Application.Common.Mappings.MappingProfile
CleanArchitecture.Razor.Application.Common.Mappings.MappingProfile
This can cause configuration collisions and inconsistent mapping.
Consolidate the CreateMap calls into one profile, or set the root Advanced.AllowAdditiveTypeMapCreation configuration value to 'true'.
```
Can there be a good solution?

janosorcsik:
Hi @Eagleleader and @neozhu!

I don't need automatic mapping profiles. So I use _IMapFrom_ or _IMapto_ or both.
_IMapto_ is the same as _IMapFrom_, but the order of parameters is different.
```csharp
public interface IMapTo<T>
{
  void Mapping(Profile profile) => profile.CreateMap(GetType(), typeof(T));
}
```

When I first use _IMapTo_ got the same error. Because I wrote two registrations, one for _IMapFrom_ and for _IMapTo_.

With the new registration in the pull request, everything works.

neozhu:
@janosorcsik thanks for the response, I did it.
image

neozhu:
image
The problem remains with DuplicateTypeMapConfigurationException 

```
public class MappingProfile : Profile
{
    public MappingProfile()
    {
        ApplyMappingsFromAssembly(Assembly.GetExecutingAssembly());
    }

    private void ApplyMappingsFromAssembly(Assembly assembly)
    {
        var mapFromType = typeof(IMapFrom<>);
        var mappingMethodName = nameof(IMapFrom<object>.Mapping);
        bool HasInterface(Type t) => t.IsGenericType && t.GetGenericTypeDefinition() == mapFromType;
        var types = assembly.GetExportedTypes().Where(t => t.GetInterfaces().Any(HasInterface)).ToList();
        var argumentTypes = new Type[] { typeof(Profile) };
        foreach (var type in types)
        {
            var instance = Activator.CreateInstance(type);
            var methodInfo = type.GetMethod(mappingMethodName);
            if (methodInfo != null)
            {
                methodInfo.Invoke(instance, new object[] { this });
            }
            else
            {
                var interfaces = type.GetInterfaces().Where(HasInterface).ToList();
                if (interfaces.Count > 0)
                {
                    foreach (var @interface in interfaces)
                    {
                        var interfaceMethodInfo = @interface.GetMethod(mappingMethodName, argumentTypes);
                        interfaceMethodInfo?.Invoke(instance, new object[] { this });
                    }
                }
            }

        }
    }
}

jasontaylordev:
Thanks, I've merged this in. 😀

PR #432, opened 2021-08-20
CICD - make it easy to see which tests passed/failed

wicksipedia:
It would be great to see which tests passed/failed in the GH workflow


wicksipedia:
needs more work since PR head branches have a read only token and can't upload test results
https://github.com/dorny/test-reporter#recommended-setup-for-public-repositories

PR #446, opened 2021-09-11
Simplify method signature

janjoosse:
Since these DependencyInjection.cs methods are called on an instance of `IServiceCollection`, I think there's no need for returning that instance?

### DIFF
--- a/src/Application/DependencyInjection.cs
+++ b/src/Application/DependencyInjection.cs
@@ -9,7 +9,7 @@ namespace CleanArchitecture.Application
 {
     public static class DependencyInjection
     {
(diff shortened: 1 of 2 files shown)


ShreyasJejurkar:
That has been done intentionally for chaining.

Right now - 
```csharp
public static void ConfigureServices(this IServiceCollection services)
{
    services.AddApplication().AddInfrastructure();
}
``` 

After your change - 
```csharp
public static void ConfigureServices(this IServiceCollection services)
{
    services.AddApplication();
    services.AddInfrastructure();
}
``` 
From my point of view, we should not change this behavior as this is breaking change for all who are using this template. And also this change not also adding much value!

PR #463, opened 2021-10-06
Reduce code duplication in configuration classes

vmutlu:
The process of calling these classes in the context class by reducing the repetition of code in the Configuration classes

### DIFF
--- a/src/Domain/Common/CommonEntity.cs
+++ b/src/Domain/Common/CommonEntity.cs
@@ -1,0 +1,8 @@
+﻿namespace CleanArchitecture.Domain.Common
+{
+    public class CommonEntity : AuditableEntity
+    {
+        public int Id { get; set; }
+        public string Title { get; set; }
+    }
+}
--- a/src/Infrastructure/Persistence/ApplicationDbContext.cs
+++ b/src/Infrastructure/Persistence/ApplicationDbContext.cs
@@ -2,6 +2,7 @@
 using CleanArchitecture.Domain.Common;
 using CleanArchitecture.Domain.Entities;
 using CleanArchitecture.Infrastructure.Identity;
+using CleanArchitecture.Infrastructure.Persistence.Configurations;
 using IdentityServer4.EntityFramework.Options;
 using Microsoft.AspNetCore.ApiAuthorization.IdentityServer;
 using Microsoft.EntityFrameworkCore;
@@ -63,6 +64,10 @@ namespace CleanArchitecture.Infrastructure.Persistence
         protected override void OnModelCreating(ModelBuilder builder)
         {
             builder.ApplyConfigurationsFromAssembly(Assembly.GetExecutingAssembly());
+
+            builder.ApplyConfiguration(new TodoListConfiguration());
+
+            builder.ApplyConfiguration(new TodoItemConfiguration());
 
             base.OnModelCreating(builder);
         }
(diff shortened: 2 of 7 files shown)


src/Domain/Common/CommonEntity.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace CleanArchitecture.Domain.Common
+{
+    public class CommonEntity : AuditableEntity
```
jasontaylordev, line 3: I think it would be better to call the class BaseEntity, as it's more of a convention - e.g. https://github.com/dotnet-architecture/eShopOnWeb/blob/main/src/ApplicationCore/Entities/BaseEntity.cs

src/Domain/Common/CommonEntity.cs
```diff
@@ -0,0 +1,8 @@
...
+    {
+        public int Id { get; set; }
+        public string Title { get; set; }
```
jasontaylordev, line 6: I wouldn't add a property such as `Title` since it wouldn't be used by all entities in most systems.

src/Domain/Common/CommonEntity.cs
jasontaylordev: It's not always going to be the case that all entities are auditable entities

src/Infrastructure/Persistence/ApplicationDbContext.cs
```diff
@@ -64,6 +65,10 @@ protected override void OnModelCreating(ModelBuilder builder)
         {
             builder.ApplyConfigurationsFromAssembly(Assembly.GetExecutingAssembly());
 
+            builder.ApplyConfiguration(new TodoListConfiguration());
```
jasontaylordev, line 68: Not necessary to apply configurations individually, since they will be applied automatically with:
            builder.ApplyConfigurationsFromAssembly(Assembly.GetExecutingAssembly());

src/Infrastructure/Persistence/Configurations/BaseConfiguration.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
+namespace CleanArchitecture.Infrastructure.Persistence.Configurations
+{
+    public class BaseConfiguration<TEntity> : IEntityTypeConfiguration<TEntity> where TEntity : CommonEntity
+    {
+        public virtual void Configure(EntityTypeBuilder<TEntity> builder)
+        {
+           // builder.Property<int>(x => x.Id).HasColumnName(@"Id").IsRequired(true).UseIdentityColumn().ValueGeneratedOnAdd();
```
jasontaylordev, line 11: Better not to check in commented out code, it doesn't typically add value and can cause confusion. Why is this here? Is it important? Can I remove it or does someone else need it?

src/Infrastructure/Persistence/Configurations/TodoItemConfiguration.cs
```diff
@@ -1,18 +1,15 @@
...
 
 namespace CleanArchitecture.Infrastructure.Persistence.Configurations
 {
-    public class TodoItemConfiguration : IEntityTypeConfiguration<TodoItem>
+    public class TodoItemConfiguration : BaseConfiguration<TodoItem>
     {
-        public void Configure(EntityTypeBuilder<TodoItem> builder)
+        public override void Configure(EntityTypeBuilder<TodoItem> builder)
         {
             builder.Ignore(e => e.DomainEvents);
```
jasontaylordev, line 10: This configuration wouldn't be required, except for DomainEvents. It might be good to move domain events to a BaseEntity. Something to consider.

jasontaylordev (commented):
Thanks for your contribution, I've provided some feedback. I hope you find it useful.
At this stage I'm not looking to bring in a BaseEntity, but it's certainly a great approach and useful in many solutions.
Happy coding!

PR #516, opened 2021-12-07
Command & Queries to use records

misha130:
Based on the PR #348 on issues #347 and the discussion in #314

This PR only applies to the commands and queries and does not change anything else.

If a command/query has a single argument then it has a constructor, otherwise it has a list of properties with get; init;


jasontaylordev:
Thank you

PR #600, opened 2022-03-31
Remove a commented line that leads to an error

NourNafea:
Remove the // from line 119 in the Sturtup.cs file


NourNafea:
opened by fault

PR #601, opened 2022-03-31
Update Startup.cs

NourNafea:
Deleted an unnecessary comment that leads to an error when running the application


jasontaylordev:
It's comment out so that ng can run independently as per; https://docs.microsoft.com/en-us/aspnet/core/client-side/spa/angular?view=aspnetcore-5.0&tabs=visual-studio#run-ng-serve-independently-1

Run the application following these steps; https://github.com/jasontaylordev/CleanArchitecture#getting-started.

PR #602, opened 2022-04-03
replace if null for null coalescing operator

VictorRH:
replace `if `validation nulls for null-coalescing operator  `??` and remove line breaks

``` c#
var entity = await _context.TodoLists
            .Where(l => l.Id == request.Id)
            .SingleOrDefaultAsync(cancellationToken);

        if (entity == null)
        {
            throw new NotFoundException(nameof(TodoList), request.Id);
        }
```

new changes 

``` c#
var entity = await _context.TodoLists
            .Where(l => l.Id == request.Id)
            .SingleOrDefaultAsync(cancellationToken) ??
            throw new NotFoundException(nameof(TodoList), request.Id);
```


jasontaylordev:
Thanks, however I prefer the clarity of the current approach.
I might clean it up further with https://github.com/ardalis/GuardClauses - check it out

PR #603, opened 2022-04-05
refactor: use generic Created/Deleted event handlers

bacongobbler:
While in this example there is only one Created/Deleted event and one Created/Deleted event handler, in a project with multiple entities emitting these events it makes sense to implement a more generic Created/Deleted event handler.

I also moved the event handlers to common as it can be re-used to log any created/deleted events.

Signed-off-by: Matthew Fisher <matt.fisher@fermyon.com>


neozhu:
thank you!

bacongobbler:
I’m not quite sure I understand the question. Could you please elaborate more on that thought? What are you trying to do with those entities?

bacongobbler:
If you’re referring to entity-specific event handlers, you could use the following function signature:

```
public class TodoItemEventHandler : INotificationHandler<DomainEventNotification<CreatedEvent<TodoItem>>>
```

You could then put that under the `CleanArchitechture.Application.TodoItem.EventHandlers` namespace as before.

Is that what you are referring to, or were you referring to something else?

neozhu:
Nice

bacongobbler:
I'd hold off on merging this one FYI... I just discovered an issue with this design pattern due to the way that aspnetcore resolves generic services. A fix is planned for .NET 7.

https://github.com/jbogard/MediatR.Extensions.Microsoft.DependencyInjection/issues/118
https://github.com/dotnet/runtime/issues/65145

neozhu:
I have a new problem
NotificationHandler gets called twice #718

neozhu:
image
Createdeventhanler not working

neozhu:
image
this code will be invoke twice
why

bacongobbler:
there’s a bug in .NET causing this behaviour. Please go read through the issues in my last comment.

https://github.com/jasontaylordev/CleanArchitecture/pull/603#issuecomment-1090795153

neozhu:
> 

Thank you for your prompt reply, I hope it can be repaired soon

jasontaylordev:
Thanks @bacongobbler, this will be a great improvement once the bug is fixed.

jasontaylordev:
Hey @bacongobbler. Thanks for your contribution however I've decided not to merge this PR in the interest of keeping the solution as simple as possible.

PR #617, opened 2022-05-02
Added .AsNoTracking() for list projections

ramax495:
As it is used CQRS pattern for the better performance in read model we could use no-tracking query by default in PaginatedListAsync and ProjectToListAsync methods in mapping extensions.


jasontaylordev:
Thanks

PR #628, opened 2022-05-20
Use existing extension for returning list

ramax495:
Use existing IQueryable extension (MappingExtensions.ProjectToListAsync) for returning list in GetTodosQueryHandler.


jasontaylordev:
Thanks, however this doesn't really add value.

PR #638, opened 2022-05-28
add enum source generator 

EngRajabi:
I added to the Enum Source Generator project that works with enums in Compile-time and is very, very fast.
[Project link](https://github.com/EngRajabi/Enum.Source.Generator)

Supernova Enum Generators


jasontaylordev:
This does not improve the solution or fix a bug. Closing.

PR #643, opened 2022-06-03
Run Acceptance Tests After Successful Deploy

jasontaylordev:
Created a new workflow to run the acceptance tests after a successful deploy to staging or prod.


jasontaylordev:
What do you think @matt-goldman?

matt-goldman:
> What do you think @matt-goldman?

Looks good!

PR #665, opened 2022-06-30
DomainEvents created as a result of an EventHandler should also be published

deceptiveSimplicity:
I think it should be possible for additional DomainEvents to occur while processing DomainEvents. Currently there is only one iteration over the list of events, so additional DomainEvents that occur are not published.
This PR is trying to change that behavior.


neozhu:
I think all domain events should be published after saving changes

PR #689, opened 2022-08-16
Using new extension method for struct values

kibernetik542:
Idea is instead of all time using chain repeated helper functions from FluentValidation using new extension methods, and advantage is does not matter int, decimal, float or double we can use for all of them, and it is also nullable version.

### DIFF
--- a/src/Application/Common/Validations/ValidationsExtensions.cs
+++ b/src/Application/Common/Validations/ValidationsExtensions.cs
@@ -1,0 +1,12 @@
+using FluentValidation;
+
+namespace CleanArchitecture.Application.Common.Validations;
+
+public static class ValidationsExtensions
+{
+    // When case for nullable structs
+    public static IRuleBuilderOptions<T, TProperty?> RangeWithWhen<T, TProperty>(this IRuleBuilder<T, TProperty?> ruleBuilder, string entity, TProperty min, TProperty max, Func<T, bool> expression) where TProperty : struct, IComparable<TProperty>, IComparable => ruleBuilder.InclusiveBetween(min,max).WithMessage($"{entity} should be in range {min} - {max}").When(expression);
+
+    // When case for non nullable structs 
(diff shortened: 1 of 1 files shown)


neozhu:
how to use it, please give me a sample

kibernetik542:
> how to use it, please give me a sample

```
public class Test
{
    public decimal? Price { get; set ; }
    public int? Quantity { get; set ; }
    public bool Enabled { get; set; }
    public double Sale { get; set; }
}
```

```
public class TestValidator : AbstractValidator<Test>
{
   public TestValidator()
   {
        RuleFor(x => x.Price).RangeWithWhen("Price", 1, 300, c => c.Price is not null);
        RuleFor(x => x.Quantity).RangeWithWhen("Quantity", 1,10, c => c.Enabled);
        RuleFor(x => x.Sale).RangeWithWhen("Sale", 10, 30, c => !c.Enabled);
   }
}
```

In my extension it is When condition, but for sure we can create however we want, advantage of this we are using for all struct based values same extension method

neozhu:
> 
I got, Thanks

IvanAntipov:
Does it make sense to propose this extension to FluentValidation repository? 
As far as I understand AIM of  CleanArchitecture repository is to provide a template with a skeleton project with a best-practice structure, for learning and/or demo and/or prototyping purposes.

In my opinion, this extension will increase, a little bit the learning curve, especially for people who are not familiar with `FluentValidation`, because this introduces an additional level in the call chain. 
 
I agree that `RangeWithWhen ` is simpler to read than `.InclusiveBetween(..).With Message().When()`, but I would prefer to have it out of the box with `FluentValidation` package.

kibernetik542:
In the end it is nothing new for FluentValidation, just extension method with ready chain methods. 
As it is FluentValidation usage in Clean Architecture skeleton. I think it makes sense propose handy extension methods. 

I can give one example which is automapper. It is extension method using ProjectTo for pagination and reflection for IMapFrom in this repository. 

IMHO related packages can be simplified witg suggestions.

jasontaylordev:
HI @kibernetik542. Thanks for the PR however I've decided not to merge in the interest of keeping the template as simple as possible. Thanks anyway and happy coding! 😀

Issue #700, opened 2022-09-13
UnhandledExceptionBehavior Gets Executed Even for Handled Exceptions


fedeantuna:
**Describe the bug**
I'm not 100% sure to call this a bug, but MediatR allows to use [Exception Handling](https://github.com/jbogard/MediatR/wiki#exceptions-handling). The problem is that by having the UnhandledExceptionBehavior as an IPipelineBehavior it becomes impossible (or at least very hard and overcomplicated) to have handled exceptions because the code under UnhandledExceptionBehavior will execute always and before the code for a handled exception.

**To Reproduce**
1. Create a command that can throw an exception
2. Create an ExceptionHandler for that particular exception and that particular command
3. Make the Exception happen
4. The UnhandledExceptionBehavior will be executed

**Expected behavior**
UnhandledExceptionBehavior is not exectued, only the Handler for the exception is executed

**Example of ExceptionHandler**
The DeleteTodoItemCommandHandler can throw the `NotFoundException`. Imagine that we want to catch that exception to log it and then consider it handled, not throw it. It might not be the best approach in a real app, but just for the sake of the example let's assume that it's really what we want. Then we can create the following ExceptionHandler:

```csharp
public class DeleteTodoItemNotFoundExceptionHandler : IRequestExceptionHandler<DeleteTodoItemCommand, Unit, NotFoundException>
{
    private readonly ILogger _logger;
    
    public DeleteTodoItemNotFoundExceptionHandler(ILogger<DeleteTodoItemNotFoundExceptionHandler> logger)
    {
        _logger = logger;
    }
    
    public Task Handle(DeleteTodoItemCommand request,
        NotFoundException exception,
        RequestExceptionHandlerState<Unit> state,
        CancellationToken cancellationToken)
    {
        var requestName = request.GetType().Name;
        
        _logger.LogWarning(exception,
            "CleanArchitecture Request: Handled Exception for Request {Name} {@Request}",
            requestName, request);
        
        state.SetHandled(default);
        
        return Task.CompletedTask;
    }
}
```

The problem is that even by doing this, the `UnhandledExceptionBehavio` will be invoked before this handler.

I found that a solution to this approach is to change the `IPipelineBehavior` from the `UnhandledExceptionBehavior` to `IRequestExceptionAction` and remove it from the registration in `ConfigureServices`. This will make the `UnhandledExceptionBehavior` run only when we have an exception that wasn't handled before.

[...]

PR #701, opened 2022-09-13
UnhandledExceptionBehavior Gets Executed Even for Handled Exceptions

fedeantuna:
This PR aims to fix the issue #700

I'm not 100% sure to call this a bug, but MediatR allows to use [Exception Handling](https://github.com/jbogard/MediatR/wiki#exceptions-handling). The problem is that by having the UnhandledExceptionBehavior as an IPipelineBehavior it becomes impossible (or at least very hard and overcomplicated) to have handled exceptions because the code under UnhandledExceptionBehavior will execute always and before the code for a handled exception.

### DIFF
--- a/src/Application/Common/Behaviours/UnhandledExceptionBehaviour.cs
+++ b/src/Application/Common/Behaviours/UnhandledExceptionBehaviour.cs
@@ -1,9 +1,11 @@
-﻿using MediatR;
+﻿using MediatR.Pipeline;
 using Microsoft.Extensions.Logging;
 
 namespace CleanArchitecture.Application.Common.Behaviours;
 
-public class UnhandledExceptionBehaviour<TRequest, TResponse> : IPipelineBehavior<TRequest, TResponse> where TRequest : notnull
+public class UnhandledExceptionBehaviour<TRequest, TException> : IRequestExceptionAction<TRequest, TException>
+    where TRequest : notnull
+    where TException : Exception
 {
     private readonly ILogger<TRequest> _logger;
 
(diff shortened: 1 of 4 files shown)


fedeantuna:
#700

jasontaylordev:
You can order the behaviours differently to produce different results. The current order is as follows:

```csharp
        services.AddTransient(typeof(IPipelineBehavior<,>), typeof(UnhandledExceptionBehaviour<,>));
        services.AddTransient(typeof(IPipelineBehavior<,>), typeof(AuthorizationBehaviour<,>));
        services.AddTransient(typeof(IPipelineBehavior<,>), typeof(ValidationBehaviour<,>));
        services.AddTransient(typeof(IPipelineBehavior<,>), typeof(PerformanceBehaviour<,>));
```

This means that the `UnhandledExceptionBehaviour` will run first, catching, logging, and rethrowing any exceptions. This is the desired behaviour. Issues with validation and not found and so on are not expected to occur since the client provides validation etc.

If you changed the order as follows:

```csharp
        services.AddTransient(typeof(IPipelineBehavior<,>), typeof(ValidationBehaviour<,>));
        services.AddTransient(typeof(IPipelineBehavior<,>), typeof(UnhandledExceptionBehaviour<,>));
        services.AddTransient(typeof(IPipelineBehavior<,>), typeof(PerformanceBehaviour<,>));
        services.AddTransient(typeof(IPipelineBehavior<,>), typeof(AuthorizationBehaviour<,>));
```

The `ValidationBehaviour` will run first, and validation exceptions will not be caught.

Be sure to access whether or not these are truly exceptional circumstances for your system. If validation / not found exceptions are normal and expected, then don't throw an exception - handle it with logic and return a relevant result indicating failure.

Hope this helps, happy coding! 😀

PR #723, opened 2022-10-27
tiny readme fix

brendan-nobadthing:
tiny fix to readme - I found the cut-paste install line didn't work for me without "--install"


jasontaylordev:
Thanks Brendan, however with the release of .NET 7 the command is now `dotnet new install`

PR #755, opened 2022-12-09
Added Todo, JWT, and API links back to the nav menu

danielmackay:
UI after change: 
image


danielmackay:
I noticed some weird HTML formatting within `nav-menu-component.html`.  If that is a formatting standard you'd like to adopt, it might be worth trying to setup some prettier rules to enforce it.

Issue #104, opened 2020-04-10
[Feature request] Use fluent validation rules to define swagger schema


fmendez89:
It is not a problem, but using the fluent validation rules, the swagger schema definition would be much more detailed. Also the generated clients definitions could benefit from this.

It would be great if the solution could use the fluent validation rules to define the swagger schema using the [nswag fluent validation extension](https://github.com/zymlabs/nswag-fluentvalidation)

Another solution could be use data annotation in dtos and commands, but would be redundant

jasontaylordev:
So, after all this time, this feature is now in place. Thanks, @Lanz86. 😀

PR #785, opened 2023-01-27
[Feature request] Use fluent validation rules to define swagger schema #104

Lanz86:
Use fluent validation rules to define swagger schema #104


jasontaylordev:
Thank you, much appreciated. 😀

Issue #776, opened 2023-01-20
API only template version


danielmackay:
**As a** software developer **I would like to** be able to create an API only version of the template **so that** I can use the template to quickly spin up an API only project with minimal manual changes

**Acceptance Criteria**
1. When creating a new project from the template add the ability to pass in a new flag to indicate API only
2. When the flag is specified, none of the Angular code should be present in the template

**Notes**
- To do this cleanly Angular may need to be separated from the API

jasontaylordev:
Covered by #786

PR #786, opened 2023-01-29
✨ Implement API only option in template

Lanz86:
API only template version #776
implemented flag apiOnly

the command created solution with Angular

<code>
dotnet new ca-sln
</code>

Create solution without Angular
<code>
dotnet new ca-sln --apiOnly=true 
</code>

### DIFF
--- a/.template.config/template.json
+++ b/.template.config/template.json
@@ -18,8 +18,17 @@
   },
   "sourceName": "CleanArchitecture",
   "preferNameDirectory": true,
+  "symbols":{
+    "apiOnly":{ "type": "parameter", "dataType":"bool", "defaultValue": "false" }
+  },
   "sources": [
     {
+      "modifiers": [
+        { 
+          "condition": "(apiOnly)",
+          "exclude": ["src/WebUI/ClientApp/**/*"]
+        }
+      ],
       "source": "./",
       "target": "./",
       "exclude": [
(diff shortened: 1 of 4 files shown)


JohanHeyvaert:
I suppose the nswag typescript client generation needs to be deactivated too. (if not already done)

Lanz86:
> I suppose the nswag typescript client generation needs to be deactivated too. (if not already done)
Yes is implemented you find directive in src/WebUI/nswag.json line 63 - 128

jasontaylordev:
@Lanz86 thank you for the contribution. The updated template supports both Angular and React, but not API only. I think we are now in a good position to implement API only too. The approach will need to change. If you look at the new [template.json](https://github.com/jasontaylordev/CleanArchitecture/blob/main/.template.config/template.json) you can see the following parameter:

```json
    "clientFramework": {
      "type": "parameter",
      "datatype": "choice",
      "choices": [
        {
          "choice": "angular",
          "description": "Use Angular"
        },
        {
          "choice": "react",
          "description": "Use React"
        }
      ],
      "defaultValue": "angular",
      "description": "The type of client framework to use"
    },
```

I think API only can be implemented by adding the choice **none**, for example:

```json
        {
          "choice": "none",
          "description": "Use API only"
        },
```

What do you think? Do you want to tackle this change?

Lanz86:
Ok @jasontaylordev sounds good. 
I will take care of it.

Lanz86:
Closed this PR and created new to support a none choice in clientFramework symbols https://github.com/jasontaylordev/CleanArchitecture/pull/886

PR #819, opened 2023-03-06
Simplified Statements 

walidaslam7:
The commit contains simplification of conditional, using and object statements


jasontaylordev:
Hi @walidaslam7. Thanks for your contribution, however, these statements have been replaced with guard clauses within #878.

PR #826, opened 2023-03-14
Transform to generic type Id in BaseAuditableEntity and BaseEntity

Lanz86:
Transform to generic type Id in BaseAuditableEntity and BaseEntity

### DIFF
--- a/src/Domain/Common/BaseAuditableEntity.cs
+++ b/src/Domain/Common/BaseAuditableEntity.cs
@@ -1,6 +1,8 @@
-﻿namespace CleanArchitecture.Domain.Common;
+﻿using CleanArchitecture.Domain.Common.Contracts;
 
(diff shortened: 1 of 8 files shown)


jasontaylordev:
Hi Antonio. The goal of the template has always been the simplest approach to enterprise application development with Clean Architecture and .NET. We should therefore focus on the general case, and let developers handle the special cases. I propose we keep `BaseEntity` simple. Perhaps we can follow the example of [eShopOnWeb](https://github.com/dotnet-architecture/eShopOnWeb) and simply add a comment; see [BaseEntity.cs](https://github.com/dotnet-architecture/eShopOnWeb/blob/main/src/ApplicationCore/Entities/BaseEntity.cs).

What do you think?

jasontaylordev:
Thanks for the suggestion. In the interests of keeping it simple I have added the comment in #873.

Issue #836, opened 2023-04-26
Allow `IMapFrom` to work with classes that don't implement a default constructor


JerrettDavis:
**Describe the bug**
If you attempt to use the `IMapFrom<TSource>`  interface on classes or records that do not have a default constructor, the built-in `MappingProfile` will throw an exception when it tries to create an instance here:

https://github.com/jasontaylordev/CleanArchitecture/blob/main/src/Application/Common/Mappings/MappingProfile.cs#L27

**To Reproduce**
1. Create a class or record without a default constructor
```csharp
public class TestClass
{
    public int Id { get; set; } 
    public string Name { get; set; }
}

public record TestClassDto (int Id, string Name) : IMapFrom<TestClass>>;
```
3. Use automapper to map `TestClass` to `TestclassDto`
```csharp
var testClass = new TestClass { Id = 1, Name = "Foo" };
var result = _mapper.Map<TestClassDto>(testClass); // <- Will error here today
```

**Expected behavior**
The expectation is that `AutoMapper`'s existing constructor mapping would be used.

--------------------------------------

In my use cases, I've replaced the `var instance = Activator.CreateInstance(type);` line with `var instance = System.Runtime.Serialization.FormatterServices.GetUninitializedObject(type);`. 

While `GetUnitializaedObject` should be used with EXTREME caution, it appears to be a perfect fit here.

jasontaylordev:
Hey @JerrettDavis, what do you think of #871?

JerrettDavis:
> Hey @JerrettDavis, what do you think of #871?

This new approach feels much more in-line with AutoMapper's suggest usage guidelines! I like it.

PR #837, opened 2023-04-26
Update `MappingProfile` to allow for `IMapFrom` implementations without a default constructor

JerrettDavis:
Per issue #836, I've updated the `MappingProfile` to leverage `FormatterServices.GetUninitializedObject` instead of `Activator.CreateInstance`. This change will allow classes without default constructors to still use the `IMapFrom` interface to automatically configure AutoMapper profiles.


jasontaylordev:
Thank you for the PR, however, the template no longer uses `IMapFrom` or the associated `MappingProfile`. Removed in PR #871

PR #868, opened 2023-06-22
Fix, maxLength bug

### LINKED ISSUE #788: WebUI - Angular - ErrorValidation
https://github.com/jasontaylordev/CleanArchitecture/blob/10ac3077707e79aaeac60a93d8863bcb109054fc/src/WebUI/ClientApp/src/app/todo/todo.component.ts#L78

As I was looking into the code and "playing" around with all its aspects, I found this little bug. As I was going to create a new List with bad title (example: More than 200 characters), the controller returned me the correct error but it didn't display in the screen. I corrected the problem to my local solution as:

`const errors = JSON.parse(error.response).errors;`

Is it something worth mentioning/changing it in a PR or have you tracked it down?

tech-gian:
Closes #788 

It doesn't allow user to write more than 200 characters, as the validation in backend. The request, also, doesn't break when an error occurs.

It's compatible with the main branch, but to be sure, I opened this PR for the .net7.0 branch. If you think it's ok, we can merge these changes to the main too.

### DIFF
--- a/src/WebUI/ClientApp/src/app/todo/todo.component.html
+++ b/src/WebUI/ClientApp/src/app/todo/todo.component.html
@@ -46,7 +46,7 @@
             <div class="flex-fill">
               <input id="{{ 'itemTitle' + i }}" *ngIf="item == selectedItem" type="text"
                 class="form-control item-input-control" [(ngModel)]="item.title" (keyup.enter)="updateItem(item, true)"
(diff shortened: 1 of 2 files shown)


tech-gian:
@jasontaylordev This is the PR to close #788.

jasontaylordev:
Thanks @tech-gian, if you don't mind - please just raise it against the main branch and I'll merge it.

PR #871, opened 2023-06-22
Simplify Mapping

jasontaylordev:
## Overview
Simplified the object-object mapping approach by removing `IMapFrom` and associated `MappingProfile`. Mappings can now be specified using a standard `AutoMapper.Profile` as nested classes or in separate files as appropriate.

## Motivation
The previous approach was becoming cumbersome to maintain and didn't align well with our project requirements. Removing `IMapFrom` and the associated `MappingProfile` will streamline the mapping process and improve code clarity.

## Changes Made
- Removed `IMapFrom` and associated `MappingProfile`.
- Introduced the use of `AutoMapper.Profile` as nested classes or in separate files for specifying mappings.
- Updated the usage of AutoMapper to closely follow the [AutoMapper Usage Guidelines](https://jimmybogard.com/automapper-usage-guidelines/).

## Example
Here's an example of the new mapping approach:

```csharp
public class TodoItemDto
{
    public int Id { get; init; }

    public string? Title { get; init; }

    public bool Done { get; init; }

    public int Priority { get; init; }

    private class Mapping : Profile
    {
        public Mapping()
        {
            CreateMap<TodoItem, TodoItemDto>().ForMember(d => d.Priority, 
                opt => opt.MapFrom(s => (int)s.Priority));
        }
    }
}
```


acesteb:
If I'm not mistaken, Mapping profile was only being referenced by the tests? i.e Using MappingProfile for DI registrations was removed in a previous PR.  If so, that feels like a low hanging fruit with low impact.

Removing IMapFrom in favor of the nested class looks looks like the bigger impact and a step in the right direction imo.  Long gone are the days of IHaveCustomMapping and IMapFrom.  They served well 😅

I guess the only "downside" is the need to explicitly call 'CreateMap' even if there are no custom mappings

jasontaylordev:
> I guess the only "downside" is the need to explicitly call 'CreateMap' even if there are no custom mappings

Agreed, it was nice to just specify the interface.

PR #882, opened 2023-06-28
Fix update entry in AuditableEntity Interceptor

MMF:
The condition "entry.State == EntityState.Added" is repeated while checking the EntityState for Add and Update.

image


jasontaylordev:
Thanks for the contribution, however, this approach is by design.
Last Modified is set on initial creation and subsequent updates.
It avoids writing code such as; `var lastModified = entity.LastModified ?? entity.Created;`

PR #884, opened 2023-06-29
Replaced DateTimeService with System.TimeProvider

tonven:
Starting from .NET 8 there is an inbuild TimeProvider Class which can be used as an abstractrion over DateTime. https://learn.microsoft.com/en-us/dotnet/api/system.timeprovider?view=net-8.0

### DIFF
--- a/src/Application/Common/Interfaces/IDateTime.cs
+++ b/src/Application/Common/Interfaces/IDateTime.cs
@@ -1,6 +1,0 @@
-﻿namespace CleanArchitecture.Application.Common.Interfaces;
-
-public interface IDateTime
-{
-    DateTime Now { get; }
-}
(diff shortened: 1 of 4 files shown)


neozhu:
I think it's better to save in UTC time, and convert to local time for display when outputting.

Tonvengo:
> I think it's better to save in UTC time, and convert to local time for display when outputting.

I agree. I was also thinking about changing Created and LastModified to DateTimeOffset and saving it to UTC time.

@jasontaylordev what do you think?

jasontaylordev:
> > I think it's better to save in UTC time, and convert to local time for display when outputting.
> 
> I agree. I was also thinking about changing Created and LastModified to DateTimeOffset and saving it to UTC time.
> 
> @jasontaylordev what do you think?

I agree, thanks for the contribution. 🙂

jasontaylordev:
Thanks again @tonven. Once the final changes come through, I'll review and merge.

rlarno:
If you change to using UTC, I suggest to also reflect that to the property names `CreatedUTC` and `LastModifiedUTC`

What these values need to contain so depend on the usecase. If these properties are for 'operational' reasons, UTC is the best value, as you can correlate typically with server logs which also (should) have UTC. If these values are to be used to define when the user created/modified the TodoItem, a full `DateTimeOffset` should be used, unless you are sure to always be operating in a single timezone.

jasontaylordev:
Closing due to inactivity, feel free to raise again.

PR #896, opened 2023-07-05
✨ Switch to Minimal API

jasontaylordev:
## Overview
Minimal API is a new way of building ASP.NET Core apps. It allows developers to build web APIs by using just a few lines of code, with a focus on simplicity and ease of use. When using Minimal API, developers can define routes and handle requests without the use of controllers or actions. Instead, the whole app can be defined in a single file.

Minimal API does take a different approach to routing when compared to ASP.NET Core controllers. While traditional controllers rely on conventions and attribute-based routing, Minimal API requires explicit route configuration. I much prefer a convention-based approach, so in this implementation I've extended Minimal API to define routes based on class and method names. This has resulted in a simple and clean implementation of Minimal API within the template.

The original PR #720 includes more in-depth information for those interested in understanding the final design and additional insights related to the changes made. Thanks to everyone who contributed to that discussion. ✨

## Changes Made

- Switched from ASP.NET Core Controllers to Minimal API
- Extended Minimal API to support convention-based routing
- Extended Minimal API to support automatic registration of endpoints
- Defined numerous defaults for endpoints including endpoint names and group names, tags, exception filter, and Open API support.
- Misc. template updates to support new approach, such as changes to generated TypeScript client.

## Example

Here's an example of how to use Minimal API within the template:

```csharp
using CleanArchitecture.Application.WeatherForecasts.Queries.GetWeatherForecasts;

namespace CleanArchitecture.WebUI.Endpoints;

public class WeatherForecasts : EndpointGroupBase
{
    public override void Map(WebApplication app)
    {
        app.MapGroup(this)
            .RequireAuthorization()
            .MapGet(GetWeatherForecasts);
    }

    public async Task<IEnumerable<WeatherForecast>> GetWeatherForecasts(ISender sender)
    {
        return await sender.Send(new GetWeatherForecastsQuery());
    }
}
```

The above example demonstrates that groups of endpoints will be defined in a single file. The endpoint group name will be based on the class name `WeatherForecasts` and the endpoint name will be based on the method name `GetWeatherForecasts`. These names will be used for routing, tags, and within the Open API specification automatically.

These classes (based on `EndpointGroupBase`) can be registered automatically within `Program.cs` as follows:

```csharp
app.MapEndpoints();
```

## Feedback
**Please review the changes and provide your feedback. Thank you! ❤️**


anguzo:
Just some food for thought.
Wanted to point out that this approach looks somewhat similar to [Carter](https://github.com/CarterCommunity/Carter). Sadly project seems kinda stale, but maybe it's worth opting into separating the "API structuring" concern from this template into a separate package? In that case [Carter](https://github.com/CarterCommunity/Carter) seems like a good starting point with the same ideology.

jasontaylordev:
Thanks @anguzo, I did look at Carter, but I wanted to align closely to ASP.NET Core Minimal API in the hopes that these or similar features will be added to ASP.NET Core. I've tried to keep the implementation very simple, without losing existing minimal API features.

Issue #898, opened 2023-07-05
Centralize constants


ChienNQuang:
In the code, role "Administrator" and policy "CanPurge" to me are like magic strings, when it comes to modify one, we have to modify all instances. And how do I know how many roles and policies I have?
My suggestion is to centralize all constants in a place, may be `src/Application/Common/Constants` like this:
``` C#
public abstract class Constants
{
   public abstract class Roles
   {
      public const string Administrator = "Administrator";
   }

   public abstract class Policies
   {
      public const string CanPurge = "CanPurge";
   }
}
```
The same goes for configuration strings in Infrastructure, but they can be addressed using Options pattern.

Any thoughts?

ramax495:
I'm using enum in my projects:

```
namespace CleanArchitecture.Domain.Enums;

public enum UserRole
{
    Administrator,
    Customer,
    Manager,
}
```
And in command and query handlers:
```
[Authorize(Roles = nameof(UserRole.Administrator))]
```

ChienNQuang:
> I'm using enum in my projects:
> 
> ```
> namespace CleanArchitecture.Domain.Enums;
> 
> public enum UserRole
> {
>     Administrator,
>     Customer,
>     Manager,
> }
> ```
> And in command and query handlers:
> ```
> [Authorize(Roles = nameof(UserRole.Administrator))]
> ```

This is indeed one way to do it. I went with class because I sort of want the constants to be able to be centralized and nested into layers. And when you call them:
`Constants.Role.Administrator`

In my project I did the same for all exception and log messages. Centralize them all into one place and effectively reuse them.

jasontaylordev:
Hey everyone. Thanks for the suggestions. Let's combine the approaches:

```csharp
namespace CleanArchitecture.Domain.Constants;

public abstract class Roles
{
    public const string Administrator = nameof(Administrator);
}

public abstract class Policies
{
    public const string CanPurge = nameof(CanPurge);
}
```

```csharp
[Authorize(Roles = Roles.Administrator)]
```

I think that removing the constants class and ensuring that the constants are within a `CleanArchitecture.Domain.Constants` namespace improves readability without losing context. The enum approach is nice, but using `nameof` reduces readability.

@ChienNQuang if you like the above approach, feel free to put together a PR and I'll push it through.

If not, let's keep the conversation going.

ChienNQuang:
> Hey everyone. Thanks for the suggestions. Let's combine the approaches:
> 
[...]

PR #911, opened 2023-07-08
Centralize role and policy constants into Domain

ChienNQuang:
Centralize roles and policies magic strings, I create separate classes in a folder/namespace to have better modularity.


jasontaylordev:
Thanks, this is looking great. 😀

#898

jasontaylordev:
Thanks, this is great! 😀

PR #920, opened 2023-07-10
Feature/nuxt

taylorc:


jasontaylordev:
Hey @taylorc - this looked interesting. Did it not work out?

taylorc:
Hey @jasontaylordev, apologies my internal workflow messed up and I accidentally created a pull request in your repo. Still an experiment but it is looking interesting. One thing holding it back is the official nuxt auth plugin is not supported for the new version of nuxt. We've looking at writing own internally but would prefer to push a fully supported auth module to this template.  If all goes well in the not too distant future we'll create a pull request.

PR #982, opened 2023-08-31
Out of the box support for master branches

konradbartecki:
Adds out of the box support for git `master` branch


jasontaylordev:
Hi Konrad, thanks for the PR however for the sake of simplicity, I'm going to keep it as main only.
For anyone that might be interested, check out [renaming the default branch from master](https://github.com/github/renaming#renaming-the-default-branch-from-master).

PR #1032, opened 2023-10-25
Use the Created DateTimeOffset for initial LastModified in MainAuditableEntityInterceptor.UpdateEntities

inghamc:
Seems more correct to have the initial LastModifiedBy timestamp exactly equal the CreatedBy timestamp rather than being very slightly later with an additional call to _dateTime.GetUtcNow()


inghamc-csu:
Refreshed (rebased) to the latest release

jasontaylordev:
Looks good, thank you. 😀

PR #1044, opened 2023-11-07
Redesigned domain events

kearns-arthur:
Simplified the domain event process.


devlear:
This seems like a step back. Aren't those events there to display in the simplest way how to use the DomainEvent system? Ideally, when implementing this, a developer would probably remove the log statement and do some business logic in the EventHandler.

kearns-arthur:
> This seems like a step back. Aren't those events there to display in the simplest way how to use the DomainEvent system? Ideally, when implementing this, a developer would probably remove the log statement and do some business logic in the EventHandler.

If that was the intent, then I completely misunderstood the purpose of the event classes.

PR #1061, opened 2023-11-25
Fix override OnModelCreating

nighttiger1990:
Follow this [post](https://learn.microsoft.com/en-us/aspnet/core/security/authentication/customize-identity-model?view=aspnetcore-2.2) and this [issue](https://github.com/dotnet/efcore/issues/9503#issuecomment-431318312)

This will fix issues related to custom TKey in Identity

> When overriding OnModelCreating, base.OnModelCreating should be called first; the overriding configuration should be called next. EF Core generally has a last-one-wins policy for configuration. For example, if the ToTable method for an entity type is called first with one table name and then again later with a different table name, the table name in the second call is used.

When trying to customize identity


KelvinSweere (approved):

jasontaylordev:
Thanks for fixing this issue. 👍

Issue #1062, opened 2023-11-26
Migration does not work properly


nighttiger1990:
**Describe the bug**
When I use any migrations command, the last pending migration migrate automatically. This cause I can't remove last migrations.

**To Reproduce**
1./ Create new project using dotnet template
```bash
dotnet new ca-sln -cf None -o YourProjectName
```
2./ Modify any entity in domain
```diff
public class TodoList : BaseAuditableEntity
{
    public string? Title { get; set; }

    public Colour Colour { get; set; } = Colour.White;
+    public string? Descriptions { get; set; }

    public IList<TodoItem> Items { get; private set; } = new List<TodoItem>();
}
```
3./ Add next migration without runing application
```bash
dotnet ef migrations add Migration1AfterInit -p .\src\Infrastructure\ -s .\src\Web\ -o Data/Migrations
```
4./ Don't run application then check database `00000000000000_InitialCreate` migrations applied (wrong from here)
5./ Modify something next
```diff
public class TodoList : BaseAuditableEntity
{
    public string? Title { get; set; }

    public Colour Colour { get; set; } = Colour.White;
    public string? Descriptions { get; set; }
+    public string? Descriptions2 { get; set; }
    public IList<TodoItem> Items { get; private set; } = new List<TodoItem>();
}
6./ Add next migration without runing application
```bash
dotnet ef migrations add Migration2AfterInit -p .\src\Infrastructure\ -s .\src\Web\ -o Data/Migrations
```
7./ Don't run application then check database `*********_Migration1AfterInit` migrations applied (wrong here too)
8./ Remove last migration (wrong here too)
```bash
dotnet ef migrations remove -p .\src\Infrastructure\ -s .\src\Web\
```
got errors
```bash
The migration '***********_Migration2AfterInit' has already been applied to the database. Revert it and try again. If the migration has been applied to other databases, consider reverting its changes using a new migration instead.
```

**Additional context**
Comment this code in `Program.cs` at `Web` Project, everything work perfectly, but we can't seed data
```c#
await app.InitialiseDatabaseAsync(); <-- Comment this ef tool will work
```

**Expected behavior**
Ef migration feature work properly with seed data.

ODumpling:
You could make it conditional by passing an arg when starting the app:

```csharp
if (app.Environment.IsDevelopment() && args.Contains("--init"))
{
    await app.InitialiseDatabaseAsync();
}
```

Then you should be able to run it using:

```bash
dotnet run --init
[...]

PR #1066, opened 2023-12-06
EF Migration cli init

ODumpling:
> ## TL;DR
> This pull request modifies the `Program.cs` file to only initialize the database in the development environment when the `--init` argument is passed.
> 
> ## What changed
> The condition to initialize the database in the development environment was changed. Previously, the database was initialized every time the application was run in the development environment. Now, it will only be initialized if the `--init` argument is passed when running the application.
> 
> A minor change was also made to remove an unnecessary line break.
> 
> ## How to test
> To test this change, run the application in the development environment with and without the `--init` argument. When the `--init` argument is passed, the database should be initialized. When it is not passed, the database should not be initialized.
> 
> ## Why make this change
> This change was made to give developers more control over when the database is initialized. This can be useful in situations where the developer does not want to reset the database every time they run the application.
</details>


ODumpling:
fixes #1062

jasontaylordev:
Thanks for the contribution. This adds a bit more complexity than the template aims for at the moment. The goal is to keep things simple, approachable, and easy to maintain. Closing this one, but really appreciate the effort and interest.

PR #1067, opened 2023-12-09
Inject ISender via EnpointGroup Constructor

Buyukcaglar:
MediatR injected once via EndpointGroup constructor and passed down to EndpointGroupBase to be utilized in Endpoint methods, rather than requiring injection in all methods. Also WebApplicationExtensions -> MapEndpoints updated to pass ISender as parameter for the constructor.


jasontaylordev:
Thank you for your contribution. For now, I will leave it as is since not all endpoint groups need to depend on `ISender`.

PR #1070, opened 2023-12-12
Replace System.Runtime.Serialization GetUninitializedObject method

max-lancaster:
Replaces obsolete `System.Runtime.Serialization` `FormatterServices.GetUninitializedObject` reference in MappingTests with equivalent version in `System.Runtime.CompilerServices`.

Based on the workaround described here:

https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0050#workaround


jasontaylordev:
Thank you. 👍

PR #1081, opened 2023-12-28
Add DataAnnotations for simplicity

aminsharifi:
Add DataAnnotations for simplicity


DennisJansenDev:
Please don't use DataAnnotations in your domain.

jasontaylordev:
For this project, we prefer using Fluent API to keep domain models separate from infrastructure. Learn more at https://learn.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/infrastructure-persistence-layer-implementation-entity-framework-core#data-annotations-versus-fluent-api.

Issue #938, opened 2023-07-17
✨ Enable Simplified Output Paths


jasontaylordev:
The upcoming release of .NET 8 introduces [simplified output paths](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#simplified-output-paths), an option to simplify the output path and folder structure for build outputs (e.g. bin and obj). The new, simplified output path structure gathers all build outputs into a common location, which makes it easier for tooling to anticipate. This can be enabled by updating **Directory.Build.props** as follows:

```diff
+ <!-- See https://aka.ms/dotnet/msbuild/customize for more details on customizing your build -->
  ﻿<Project>
      <PropertyGroup>
          <TargetFramework>net8.0</TargetFramework>
          <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+         <ArtifactsPath>$(MSBuildThisFileDirectory)artifacts</ArtifactsPath>
      </PropertyGroup>
  </Project>
```

Learn more about simplified output paths here; https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#simplified-output-paths.

Unfortunately, this feature is not compatible with **NSwag.MSBuild** so implementation will need to wait until this issue is resolved; https://github.com/RicoSuter/NSwag/issues/4485.

samabos:
Hi, it seems this issue has been resolved.

Protiguous:
Hi.
Can someone verify that this issue has been resolved?
And if so, can this issue be closed?

jasontaylordev:
This is fixed.

PR #1096, opened 2024-01-19
enable: Simplified Output Paths on .net8 #938

Samandar02:
Simplified Output Paths enabled  on .net8 #938


jasontaylordev:
Glad to see this working, thank you. 👍

PR #1105, opened 2024-02-13
Update Web-webapi.http and replaced deprecated DependencyInjection

n0one42:


n0one42:
- Updated the http file which refers to the wrong parameter (username)
- Replaced the deprecated DependencyInjection with AutoMapper

n0one42:
Duno why this should fail. Also cant see what the error should be. For me it works and I am using it atm exactly like that.

src/Application/Application.csproj
```diff
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="Ardalis.GuardClauses" />
-    <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" />
+    <PackageVersion Include="AutoMapper" />
```
jasontaylordev, line 10: This should be `<PackageReference Include="AutoMapper" />`. PackageVersion is used in **Directory.Packages.props**.

n0one42: Oh... my bad... I overlooked this. Thanks

PR #1156, opened 2024-04-01
add null forgiving operator

aitchdot:
add null forgiving operator to ITestDatabase to suppress warnings


jasontaylordev:
Thanks, however this is already done.

PR #1177, opened 2024-05-07
Fix: "EF migrations remove" command work with NSwag enabled

nighttiger1990:
Fix this issue #1062

### DIFF
--- a/src/Web/Program.cs
+++ b/src/Web/Program.cs
@@ -14,7 +14,11 @@ var app = builder.Build();
 // Configure the HTTP request pipeline.
 if (app.Environment.IsDevelopment())
 {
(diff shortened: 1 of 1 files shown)


Morlin:
I faced the same issue. Here are the breaking changes from MS https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-9.0/breaking-changes#pending-model-changes
https://github.com/dotnet/efcore/issues/35285

jasontaylordev:
Thanks for the contribution. EF Core migrations have been removed from the template in favor of a simplified initialization strategy. This PR is no longer applicable and will be closed. Appreciate the effort and interest in the project.

PR #1225, opened 2024-09-20
Update azure-dev.yml, added a name for the workflow.

southernsun:
Added a name for the workflow so it easily identified on the GitHub Actions page.

### DIFF
--- a/.github/workflows/azure-dev.yml
+++ b/.github/workflows/azure-dev.yml
@@ -1,3 +1,5 @@
+name: Azure DevOps Deployment
+
 on:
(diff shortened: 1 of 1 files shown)


jasontaylordev:
Thanks. This is not an Azure DevOps deployment, it's a GitHub Actions deployment that forms part of the solution when you run 'dotnet new ca-sln'. It's a pipeline that uses the Azure Developer CLI.

When you run ' dotnet new, ' the other pipelines won't be included, so I'll leave the name off for now.

Issue #1240, opened 2024-10-12
Domain events dispatched before save


fdugdale112:
**Describe the bug**
Domain events are dispatched from the SavingChangesAsync in DispatchDomainEventsInterceptor. If the save fails, events are still dispatched. 
I have fixed this in my application by calling DispatchDomainEvents from an override of SavedChangesAsync instead of SavingChangesAsync.

**To Reproduce**
1. in any entity configuration add an index such as:
`builder.HasIndex(t => t.PropertyName).IsUnique();`
2. Save two entities with the same property value
4. Any domain events on the second entity will still be dispatched
5. See error

**Expected behaviour**
Domain events shouldn't be dispatched when save fails

fdugdale112:
#1241

jasontaylordev:
There are pros and cons to dispatching domain events before or after saving changes. You've found one of the cons. You can read more here: https://learn.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/domain-events-design-implementation#the-deferred-approach-to-raise-and-dispatch-events. I should create an ADR for this decision.

fdugdale112:
Ah, actually I think it would be better if both scenarios were supported. ideally the event handler should be able to decide whether the effect happens before, after or a combination. I don't think there is an easy way to achieve this with the current apis. 

do you have any recommendations on how I can do this or any reason why I should avoid doing this?

PR #1241, opened 2024-10-12
Replaced SavingChangesAsync with SavedChangesAsync

fdugdale112:


jasontaylordev:
Keep existing approach as per #1240

PR #1259, opened 2024-11-11
upgrade Angular17 to 18

Samandar02:
Upgrading details
- angular base dependencies
- convert directive based conditionals and loops to built-in way


jasontaylordev:
See #1269 - needed to add some changes

Issue #1273, opened 2024-11-18
Add support for .NET Aspire


jasontaylordev:
### Discussed in https://github.com/jasontaylordev/CleanArchitecture/discussions/1056

<div type='discussions-op-text'>

<sup>Originally posted by **chrisrlewis** November 22, 2023</sup>
The most interesting news from the dotnetconf '23 for me was Aspire. I realise this is early days, and the announcement wasn't widely known about but are there plans to incorporate Aspire orchestration, deployment, etc. into this template?</div>

jasontaylordev:
Closing as complete

PR #1274, opened 2024-11-18
✨ Add support for .NET Aspire

jasontaylordev:
Added support for .NET Aspire. Create a new solution and include `--use-aspire`.
Requested in #1273


src/AppHost/Program.cs
```diff
@@ -0,0 +1,5 @@
+var builder = DistributedApplication.CreateBuilder(args);
+
+builder.AddProject<Projects.Web>("web");
+
+builder.Build().Run();
```
melotic, line 5: Should we add a SQL Server component here? The template uses a LocalDb by default.

jasontaylordev: Thanks for the suggestion. I've updated it to include SQL Server integration.

pramodkumarjaiswal: Hi, I'm trying to setup a new WebAPI solution using "--use-aspire" but it's not working, and says -is not a valid option.
Could you please suggest the right commend to create a new Aspire solution from template?

melotic: > Hi, I'm trying to setup a new WebAPI solution using "--use-aspire" but it's not working, and says -is not a valid option. Could you please suggest the right commend to create a new Aspire solution from template?

This PR is not merged and not released yet, so you'd have to build and install the templates from this branch to use the flag.

pramodkumarjaiswal: I tried below and it gave me same error. Am I missing something?

pramodkumarjaiswal: pramodjaiswal@192 CleanArchitecture % dotnet new ca-sln --use-aspire -cf None -o OrchestrationEngine
Error: Invalid option(s):
--use-aspire
   '--use-aspire' is not a valid option

jasontaylordev: Sorry, that will not work until this PR is merged and a NuGet package is published. I'll try to get it done today.

PR #1320, opened 2025-04-19
Replace FluentAssertions with Shouldly in Tests

Kamyab7:
This pull request refactors all test projects to use [Shouldly](https://docs.shouldly.org/) instead of FluentAssertions, which recently adopted a commercial license.

**Changes included:**

- Replaced all usages of FluentAssertions with equivalent Shouldly syntax.
- Removed FluentAssertions from all project references.
- Updated README to include Shouldly in the list of testing libraries.

This helps keep the project free of commercial license concerns while maintaining clear, readable test assertions.

Closes #1309 .

Issue #1206, opened 2024-07-29
Replace AutoMapper with Mapperly


neva1842:
**Is your feature request related to a problem? Please describe.**
This feature request is not related to a problem. It is an exploratory idea to consider replacing AutoMapper with Mapperly.

**Describe the solution you'd like**
I propose considering the replacement of Automapper with Mapperly in Jason Taylor's CleanArchitecture template. According to the benchmarks provided in [this GitHub repository](https://github.com/mjebrahimi/DotNet-Mappers-Benchmark), Mapperly demonstrates superior performance compared to Automapper. This enhancement can potentially lead to performance gains in our applications. Moreover, Mapperly supports IQueryable projections, which aligns well with our existing use cases.

m2017atTR:
Hi neva1842,

Automapper have support about projection  .Project() ...
[https://ojdevelops.com/2015/09/using-automapper-in-entity-framework.html](url)

fizmhd:
Hi,
We have used Mapperly  in our project and it was not a good choice.
Its a good library, when the mapping is direct, and simple mappings.

But when your property names are different, then you need to mention it by attributes.
It grows longer, when you have many properties.

When it comes to mapping Child classes, it becomes harder.

So we had to move away from Mapperly, and use simple mapping classes.

Sharing, my feedback on library, I am not against it.

Kamyab7:
Hi!

Thank you for bringing this up. In response to the idea of replacing AutoMapper with Mapperly, I've recently made a similar change in this pull request, but for slightly different reasons.

I decided to replace AutoMapper with manual mapping in the project to avoid introducing additional dependencies for a simple task like mapping. This change also came about because AutoMapper has transitioned to a commercial license, and to keep the project fully open-source and free of such dependencies, I opted for manual mappings instead.

While performance considerations are definitely important, manual mapping also offers excellent performance as it avoids the overhead that comes with reflection and the abstraction layer AutoMapper provides. It's a lightweight approach that can perform well even as the size of the data grows, and it gives us full control over the mapping logic, making it easy to optimize for specific cases if needed.

I appreciate the suggestion regarding Mapperly, and it could indeed be a good option in the future if we need specific features that it offers. However, for now, manual mapping serves our purpose well and keeps the codebase simple, clean, and performant.

Feel free to continue the discussion if you think there are other benefits to explore!

jasontaylordev:
[...]

PR #1321, opened 2025-04-19
refactor: remove AutoMapper and replace with manual mapping due to licensing changes

Kamyab7:
This PR removes AutoMapper from the application and replaces all usages with manual mapping functions.

### 🚨 Why?
AutoMapper has recently adopted a commercial license for production use. To avoid potential licensing issues and to keep the project open and dependency-light, this change removes AutoMapper entirely.

### ✅ Changes:
- Removed AutoMapper library and DI configuration
- Replaced `ProjectTo<T>()`, `.Map()` calls, and `Profile` classes with explicit mapping functions
- Added static mappers for:
  - `TodoItemBriefDto`
  - `TodoItemDto`
  - `TodoListDto`
  - `LookupDto`
- Updated query handlers and mapping extensions to support `Func<TSource, TDestination>`
- Simplified the mapping logic, making it more transparent and maintainable

### 🧩 Dependency:
This PR **depends on** [#1320](https://github.com/jasontaylordev/CleanArchitecture/pull/1320), as it introduces the `Shouldly` library used in tests.

### 📚 Housekeeping:
- Removed `AutoMapper` from the list of dependencies in the README

🔒 **Closes**: [#1206](https://github.com/jasontaylordev/CleanArchitecture/issues/1206)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -27,7 +27,6 @@
     <PackageVersion Include="OpenTelemetry.Instrumentation.Runtime" Version="1.9.0" />
     <!--#endif-->
     <PackageVersion Include="Ardalis.GuardClauses" Version="4.6.0" />
-    <PackageVersion Include="AutoMapper" Version="13.0.1" />
     <PackageVersion Include="Azure.Extensions.AspNetCore.Configuration.Secrets" Version="1.3.2" />
     <PackageVersion Include="Azure.Identity" Version="1.13.1" />
     <PackageVersion Include="coverlet.collector" Version="6.0.2" />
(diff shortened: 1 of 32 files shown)


juanmarquezruiz:
With all due respect, I believe the best decision would have been to support ourselves with other libraries that are aiming to stand out in light of this potential situation with AutoMapper. There are some well-regarded alternatives, and it would have required even less refactoring. Additionally, our friend Jimmy Bogard communicated that the current versions can still be used, but maintenance is not a priority for them.

jasontaylordev:
Thank you for your contribution, however [I'm going to continue using both AutoMapper and MediatR for the time being](https://github.com/jasontaylordev/CleanArchitecture/discussions/1346).

PR #1335, opened 2025-05-24
Changed PipelineBehavior ordering

emenychtas:
Any request validator should only be called after authorization.

Validators can impose load to the database and should be protected.


deceptiveSimplicity:
To me it looks like you moved Authorization after Validation.

juanmarquezruiz:
AuthorizationBehaviour should be before validation, because there is no point in validating if the user is not authorized.

jasontaylordev:
Thanks for the suggestion, but the current ordering is intentional. Authorization runs before validation because it is cheap (claims/roles check only) and rejects unauthorized requests before any potentially expensive validation work runs. The PR description actually argues for the current order, but the code change reverses it.

PR #1340, opened 2025-06-24
Refactor endpoint mapping to use RouteGroupBuilder

samanazadi1996:
Hi,  
I updated the endpoint mapping to use RouteGroupBuilder instead of WebApplication.  
This lets us organize routes better and apply settings more precisely—for example, requiring authorization on just one method instead of the whole group.  
I also added a way to name these groups more clearly instead of relying on class names.  

Please review when you get a chance and let me know if you have any feedback.  
Thanks!


jasontaylordev:
Thank you 👍

PR #1344, opened 2025-07-08
Refactor the architecture.

Ayush2395:
- Added 2 interfaces IDomainEvent and IAuditableEntity in the Domain, I've putted this in common folder.
- Change the Event from class to record.
- In infrastructure class library, I've changed the interceptors with IDomainEvent and IAuditableEntity.
- Added the IAuditableEntity and IDomainEvent in the ApplicationUser entity.
- These changes helps to solve the complexity of project when it is scalling, that's I've experienced.
- Move the ApplicationUser entity from Infrastructure to Domain.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -19,6 +19,7 @@
     <!--#endif-->
     <PackageVersion Include="Aspire.Hosting.AppHost" Version="$(AspireVersion)" />
     <PackageVersion Include="Microsoft.Extensions.Http.Resilience" Version="9.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Identity.Stores" Version="9.0.6" />
     <PackageVersion Include="Microsoft.Extensions.ServiceDiscovery" Version="9.0.0" />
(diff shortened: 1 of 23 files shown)


jasontaylordev:
Thanks for the contribution @Ayush2395 - there are some great changes here! Could you please update your branch with the latest main and resolve the conflicts so we can review it properly?

One note: I generally prefer not to make all entities auditable, so I’d rather have `BaseAuditableEntity` inherit from `BaseEntity`, not the other way around. Also, we can probably rename `ApplicationUser` to `User`.

jasontaylordev:
Thanks for the contribution, but I'm going to close this one.

The core problem is that moving `ApplicationUser` to Domain requires adding `Microsoft.Extensions.Identity.Stores` to the Domain project, which brings an ASP.NET Core Identity dependency into the domain layer - exactly what Clean Architecture aims to avoid.

There are a few other issues too: the inheritance hierarchy is inverted (every entity is now auditable with no opt-out), `IDomainEvent` is a misleading name for what is actually "entity that has domain events", the migration shouldn't be included in a template PR, and some template conditionals were removed from the snapshot which would break API-only builds.

PR #1348, opened 2025-07-18
use mit mediator

dzmprt:


dzmprt:
A minimal example demonstrating the use of MitMediator instead of MediatR https://github.com/dzmprt/MitMediator

PR #1349, opened 2025-07-18
Make BaseEntity generic & added default 'int' key

alitaami:
**Changes**
- Added generic BaseEntity<TKey> class with nullable Id property of type TKey.
- Updated non-generic BaseEntity class to inherit from BaseEntity<int> as a default implementation.

**Explanation**
This change introduces a generic version of the BaseEntity class to support different types of primary keys. The existing non-generic BaseEntity now inherits from BaseEntity<int>, making int the default key type to maintain backward compatibility.


jasontaylordev:
Thanks for the contribution. This adds a bit more complexity than the template aims for at the moment. The goal is to keep things simple, approachable, and easy to maintain. Closing this one, but really appreciate the effort and interest.

PR #1375, opened 2025-08-30
Update InitaliseAsync method for late development

YunusOzdemirr:


jasontaylordev:
Thanks for the contribution. The EnsureDeleted/EnsureCreated approach in InitialiseAsync is intentional for the development scenario — see https://jasontaylor.dev/ef-core-database-initialisation-strategies for the reasoning. Switching to MigrateAsync would be a significant strategy change and is better suited as a customisation in your own project.

PR #1378, opened 2025-09-01
Added missing files for build angular using build:ssr

RDB-DA:
_**Update**_
Added missing files
  - src/main.server.ts
  - tsconfig.server.json

Update angular.json to include the references of missing files
  - angular.json

**_Build SSR_**
  - install node packages using "npm i"
  - Build the SSR using "npm run build:ssr"


jasontaylordev:
Thanks for the contribution. The Angular frontend is due for a complete overhaul, with an upgrade to v21 - once that's done, SSR support can be revisited properly at that point.

PR #1384, opened 2025-09-10
fix: 'Users' does not implement inherited abstract member 'EndpointGr…

nighttiger1990:
When create new API App only using command below
```bash
dotnet new ca-sln -cf None -o <dbname> -ua -db postgresql
```

Run app in Visual Studio got error
```
'Users' does not implement inherited abstract member 'EndpointGroupBase.Map(RouteGroupBuilder)'
'Users.Map(WebApplication)': no suitable method found to override
Argument 2: cannot convert from 'ERP.LMS.Web.Endpoints.Users' to 'string'
```

### DIFF
--- a/src/Web/Endpoints/Users.cs
+++ b/src/Web/Endpoints/Users.cs
@@ -5,10 +5,9 @@ namespace CleanArchitecture.Web.Endpoints;
 
 public class Users : EndpointGroupBase
 {
-    public override void Map(WebApplication app)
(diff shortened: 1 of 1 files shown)


multi-thiagofernandes (approved):

jasontaylordev:
Thanks for the contribution! This issue has already been resolved in main — `Users.cs` already implements `Map(RouteGroupBuilder)` correctly. Closing as resolved.

PR #1403, opened 2025-12-11
Add README.md files to project roots (Fixes #1364)Docs/add readmes 1364

### LINKED ISSUE #1364: [Readme] What about adding a Readme on each of the Project roots for better getting started guidiance directly in template?
**Is your feature request related to a problem? Please describe.**

discoverd this nice template here and as I would like to learn about Clean Architecture, I am not knowing anything about those different Slices of such solutions.

**Describe the solution you'd like**

It would be great if there maybe would at least be an optional argument available, for those beginners like me, that would also add a ReadMe.md file on the root of every Project/Slice root like "Domain", "Integration" and so on, so that would tell us some valueable things like:
- What is that slice meant for -> What should we put in there, what maybe not?
- For e.g. the one including MediatR -> You find this package here for xy but as that gone under commercial licensing, you could potentially replace it with (...?) 
- "Web" for example compared to "Application" I would have a hard time to difference between their purposes. Which of them is meant to be the "regular" Programm.cs deliverer? And what is then the task for the other one?
- For what do we need Eventing on a API? - Yes, this one could mean for a expert, but I would not know of any situation, where I would have send an http request to a api and got an event or eventhandler back? so whats up with that?

**Describe alternatives you've considered**

Dont know, I currently not see much sense in that structure, while of couse, seperation of DataObjects like TodoItem from the Endpoints might be reasoned. Started rebuilding my Solution structure along this, but at the point I discovered that I do not know about a proper replace Name for that Application Project and not getting the reason for it, I came to the point that I would not be able to use that.

**Additional context**

Why this might be not usable at this point for me?
- Because e.g. if the Client App is WinUI or whatever else, the Names "Application" and "App" of any Project in my Solution would cause a naming problem for me 🤷 
- Secondary, as mentioned in the requested possible contents for the Readme files, I would miss some information that would tell me about the reasons and tasks of each of them, so that would enable me to transfer this knowledge to not just use this template as-is, then transfer this... I think it is defintly a valuable resource, if I would understand it! but to transfer it to own solutions maybe and use it as a template, a suggestion and collection of expirience from senior/midlevel developers 👍

sc2107:
This PR adds README.md files to the Domain, Application, Infrastructure, and Web project folders under /src.

Each README includes:
- Responsibilities of the layer
- Allowed dependencies
- Notes on what should and should not be included

This improves onboarding and directly addresses Issue #1364.

### DIFF
--- a/src/Web/README.md
+++ b/src/Web/README.md
@@ -1,0 +1,16 @@
+# Web Layer
+
+The **Web** project contains the user interface and HTTP endpoints of the application.
+
+## Responsibilities
+- Controllers or Minimal APIs  
+- Endpoint routing  
+- Middleware and filters  
+- Dependency injection setup for all layers  
+- Hosting the frontend (Angular/React) if included  
+
+## Dependencies
+- Depends on **Application**, **Infrastructure**, and **Domain**
+
+## Notes
+This is the entry point of the application at runtime and composes all layers together.


jasontaylordev:
Thanks for the contribution! Great addition. We noticed the Domain README listed event handlers, but those live in the Application layer — fixed in a follow-up commit.

PR #1407, opened 2025-12-19
Refactor endpoint registration to use DI

samanazadi1996:
Moved endpoint registration to DI, now can instantiate endpoints with injected services.

### DIFF
--- a/src/Web/DependencyInjection.cs
+++ b/src/Web/DependencyInjection.cs
@@ -35,6 +35,8 @@ public static class DependencyInjection
         builder.Services.Configure<ApiBehaviorOptions>(options =>
             options.SuppressModelStateInvalidFilter = true);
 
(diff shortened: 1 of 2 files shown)


jasontaylordev:
Thanks for the contribution. The idea is technically sound, but I'm going to close this one.

The endpoint group classes in this template (`TodoItems`, `TodoLists`, etc.) are purely structural - they exist only to call `groupBuilder.Map*` and have no runtime state or service dependencies. Services reach the handlers via minimal API's parameter binding, not through the endpoint group constructor. So registering them in DI enables a pattern that isn't used or demonstrated anywhere in the template.

The current `Activator.CreateInstance` approach is simpler and appropriate for what endpoint groups actually do here. Adding DI registration for a capability the template doesn't use would be complexity without a clear teaching value.

PR #1409, opened 2025-12-26
Optimize role authorization lookup to O(1) using HashSet

Kamyab7:
Optimize role authorization lookup to O(1) using HashSet for efficient role membership checks.


jasontaylordev:
Thanks for the contribution\! Users typically have very few roles so the HashSet overhead isn't worth it in practice. Keeping the simpler approach for now.

PR #1425, opened 2026-02-10
Docs: clarify backend-only development and Visual Studio errors

djadhav00003:
This PR adds a short clarification for backend-only contributors.

It explains:
- Expected frontend (Angular/TypeScript) errors when npm dependencies are not installed
- That backend builds can be validated using dotnet restore/build/test
- That frontend tooling is optional for backend-only work

This should help reduce confusion for new contributors using Visual Studio.


djadhav00003:
Added an additional functional test to ensure a NotFoundException is thrown 
when attempting to create a TodoItem with an invalid ListId.

Unit tests pass locally.
Functional test requires Docker (Testcontainers) to run.

Happy to adjust if needed.

jasontaylordev:
Thanks for the contribution\! We're planning to overhaul the way frontends are integrated, which should resolve the confusion this addresses — so holding off on README changes in this area for now. For the test in CreateTodoItemTests.cs, that looks like a good addition — please open a separate PR for that and we'll take a look.

PR #1442, opened 2026-02-22
Replace SpecFlow with Reqnroll

LukasGrafTT:
# Problem
The SpecFlow packages have all been marked as deprecated, and the source repositories unpublished. It is safe to assume the package is unmaintained.

# Proposed solution
Use the active fork called [Reqnroll](https://github.com/reqnroll/Reqnroll).

# Potential breaking change
Reqnroll does not generate a `TestExecution.json` by default. This file was used by LivingDoc to generate a HTML report, but this was not set up for this template. LivingDoc users can configure [formatters](https://docs.reqnroll.net/latest/reporting/reqnroll-formatters.html) to generate the reports.

# Housekeeping
Reqnroll introduces a flag that excludes generated code-behind files from the source code, which is [recommended practice](https://docs.reqnroll.net/latest/installation/setup-project.html).

---

Lastly, thank you for making and maintaining this awesome template!


jasontaylordev (approved):
Thank you\! This is something I've been meaning to get to for a while. Really appreciated you taking the time to put this together.

PR #1443, opened 2026-02-23
Fix: Upgrade Azure Web app to .NET 10

LukasGrafTT:
I assume this was an oversight in #1393 

For me, the web app did not launch correctly until I changed this setting and redeployed. (I also have an unrelated issue with EF Core failing to initialize the database. I am still investigating that one.)

### DIFF
--- a/infra/services/web.bicep
+++ b/infra/services/web.bicep
@@ -28,7 +28,7 @@ module appService '../core/host/appservice.bicep' = {
     applicationInsightsName: applicationInsightsName
     keyVaultName: keyVaultName
     runtimeName: 'dotnetcore'
-    runtimeVersion: '9.0'
+    runtimeVersion: '10.0'
     healthCheckPath: '/health'
     appSettings: {
       ASPNETCORE_ENVIRONMENT: 'Development'


jasontaylordev:
Nice catch @LukasGrafTT, thank you!

I suspect your database initialization issue is due to migrations not being enabled. Take a look here; https://github.com/jasontaylordev/CleanArchitecture/blob/c7aae661ca384caef83058f55ea3a6721606c00d/src/Infrastructure/Data/ApplicationDbContextInitialiser.cs#L43-L45

Happy coding! 😀

LukasGrafTT:
> Nice catch @LukasGrafTT, thank you!
> 
> I suspect your database initialization issue is due to migrations not being enabled. Take a look here;
> 
> https://github.com/jasontaylordev/CleanArchitecture/blob/c7aae661ca384caef83058f55ea3a6721606c00d/src/Infrastructure/Data/ApplicationDbContextInitialiser.cs#L43-L45
> 
> Happy coding! 😀

Yes, that was the issue. Replacing `EnsureDeletedAsync` and `EnsureCreatedAsync` with `MigrateAsync` fixed it. Makes sense to use migrations once the app is ready for deployment. Thank you!

